PR #62, opened 2019-04-15
Turn on operator logging

alex-zaitsev:


alex-zaitsev:
Apparently it is already there

PR #88, opened 2019-05-13
XML config generator rewrite && ensure user config has mandatory fileds: profile,quota,networds,password

sunsingerus:


alex-zaitsev:
Please remove default for password. Password should always be provided. In minimum possible spec it looks like:

  configuration:
    users:
      demo/password: demo

PR #107, opened 2019-05-31
Enhance SQL queries over the cluster

sunsingerus:


alex-zaitsev:
I don't like this 'retry' approach for schema creation and DNS. It especially questionable on huge clusters, since it all runs sequentially. It should be asynchronous. I would store somewhere (configmap) a state that we need to run certain task on certain pod, check periodically, and clean up the state on success

PR #117, opened 2019-06-18
zookeeper recommended settings

filimonov:
1) log4u  settings differs (rather not important),
2) I've increased the heap size to 1Gb to avoid additional costs, but default 2Gb used in kubernetes-zookeeper sound better
3) kubernetes-zookeeper image from google fixes both lower and upper size of the heap (rather it's ok).
4) preAllocSize and snapCount can't be changed in that image
https://github.com/yandex/ClickHouse/blob/master/docs/en/operations/tips.md#zookeeper


alex-zaitsev:
Fixed in https://github.com/Altinity/clickhouse-operator/commit/fd2b5233bd4186218bf2247893bfffb7d2bdbbff

PR #187, opened 2019-10-24
Add the "system_replicas_is_session_expired" metric to the exporter.

teralype:
This metric reports which replicated tables have seen their Zookeeper sessions expired. This metric is labeled by database and table, plus additional labels that are added by default (e.g. ClickHouse hostname).

Expired Zookeeper sessions cause ClickHouse to start switching tables to read/only which is likely a symptom that something is wrong with Zookeeper or the network itself. In general, when this happens, the operator needs to take action in order to fix this situation.


teralype:
@alex-zaitsev mind taking a look at this? I think this is an interesting addition to the ClickHouse operator. I've also taken the liberty to clean up the code a little bit.

PR #270, opened 2020-02-26
fix metrics-exporter goroutine leak: DB connection is created per query

guoshuai2016:
1. keep DB connection in global variable and initialize each DB connection once, instead of creating a DB connection for each query.
2. Move deadline context to clickHouseQueryScanRows, to prevent the case that Rows is closed before scan is finished.


sunsingerus:
merged into [0.9.2](https://github.com/Altinity/clickhouse-operator/tree/0.9.2)

Issue #176, opened 2019-10-16
Operator Pod Memory leak


MetalRex101:
![image](https://user-images.githubusercontent.com/17021192/66916831-c3719300-f024-11e9-9ceb-e59ce5a906a6.png)

Operator version - 0.6.0

The problem is that after operator pod reach some memory limit, it would be Evicted by kubernetes. We have to manually clear these pods to prevent them from using resources.

alex-zaitsev:
@MetalRex101 , could you advise on your Kubernetes version and ClickHouse cluster setup. We can not reproduce memory leak in our environment. Could you check if you see it growing on 'clickhouse-operator' or 'metrics-exporter' container? It can be checked with a command like:

# kubectl top pod <your_operator_pod_name> --containers=true -n <your_namespace>

MetalRex101:
Kubenetes version:
```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.3", GitCommit:"5e53fd6bc17c0dec8434817e69b04a25d8ae0ff0", GitTreeState:"clean", BuildDate:"2019-06-06T01:44:30Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"12+", GitVersion:"v1.12.9-gke.15", GitCommit:"b48a8d693e191192e27c2f807daa51b54d0b0a61", GitTreeState:"clean", BuildDate:"2019-08-12T17:49:30Z", GoVersion:"go1.10.8b4", Compiler:"gc", Platform:"linux/amd64"}
```
Operator pod top:
```
POD                                    NAME                  CPU(cores)   MEMORY(bytes)
clickhouse-operator-6b7f548688-m5857   clickhouse-operator   8m           36Mi
```

Clickhouse cluster configuration:
```
Name:         chi-clickhouse-db-common-configd
Namespace:    clickhouse
Labels:       clickhouse.altinity.com/app=chop
              clickhouse.altinity.com/chi=clickhouse-db
              clickhouse.altinity.com/chop=0.6.0
Annotations:  <none>

Data
====
remote_servers.xml:
----
<yandex>
    <remote_servers>
        <!-- User-specified clusters -->
        <default>
            <shard>
                <internal_replication>true</internal_replication>
                <replica>
                    <host>chi-clickhouse-db-default-0-0</host>
                    <port>9000</port>
                </replica>
                <replica>
[...]

PR #279, opened 2020-02-27
Add metrics exporter e2e tests

### LINKED ISSUE #239: metrics exporter lose monitored ClickHouse instances list and stop publish metrics, after restart metrics-exporter container inside clickhouse-operator pod
Steps to reproduce
- install clickhouse-operator
- install any kind: ClickhouseInstallation from `chi-examples`
- reboot metrics-exporter via
```
kubectl exec -n $( kubectl get pods --all-namespaces | grep clickhouse-operator | awk '{print $1 " " $2}') -c metrics-exporter reboot
``` 
- run port-forward to metrics exporter 8888
```
kubectl -n $( kubectl get svc --all-namespaces | grep clickhouse-operator-metrics | awk '{print $1}') port-forward service/clickhouse-operator-metrics 8888
```
- open http://localhost:8888/chi

## Expected results
json array with current running clickhouse instances

## Actual results
empty json

This state also happens in following situations:
- OOP Kill clickhouse-operator/metrics-exporter container
- reschedule clickhouse-operator/metrics-exporter 
- whole restart kubernetes node / cluster 

## Solution proposal
- simple add retries and timeout options to on `MakeRESTCall` in file https://github.com/Altinity/clickhouse-operator/blob/master/pkg/apis/metrics/rest.go
- add retries and logging to https://github.com/Altinity/clickhouse-operator/blob/master/pkg/controller/chi/controller.go on `updateWatchAsync`

Slach:
- add test_metrics_exporter.py for check regression and close #239
- add Vagrantfile environment for the ability test circle.ci scripts local before deploy
- currently all tests passed on k8s 1.14.10
- change host_regexp cause DNS PTR lookup incomind POD ip on Clickhouse side can lookup to ClusterIP service instead of headless service

look like #176  fixed on 0.9.2 branch


tests/test.py
```diff
@@ -296,103 +301,101 @@ def test_015():
...
-                kube_createns(kubectl.namespace)
-
-    #    with Module("examples", flags=TE):
-    #        examples=[test_examples01_1, test_examples01_2, test_examples02_1, test_examples02_2]
-    #        for t in examples:
-    #            run(test=t, flags=TE)
-        
+            if kube_get_count("pod", ns='--all-namespaces', label="-l app=clickhouse-operator") == 0:
+                installer = get_full_path('../deploy/operator-web-installer/clickhouse-operator-install.sh')
+                installer = shell(f"bash -x {installer}")
```
alex-zaitsev: Why do you need a shell script? You can use kubectl apply instead

Slach: fixed

PR #322, opened 2020-03-27
README: Update grafana-service (to be the same as grafana-operator name)

wiktor2200:
Grafana service name is changed when created using grafana-operator.
Here is this change in `grafana-operator` repo: https://github.com/integr8ly/grafana-operator/blob/master/pkg/controller/model/constants.go

### DIFF
--- a/deploy/grafana/grafana-manually/grafana.yaml
+++ b/deploy/grafana/grafana-manually/grafana.yaml
@@ -66,7 +66,7 @@ data:
 apiVersion: v1
 kind: Service
 metadata:
-  name: grafana
+  name: grafana-service
   labels:
     app: grafana
 spec:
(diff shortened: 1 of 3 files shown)


sunsingerus:
Ok, thanks!
We need to update Grafana's manual installation path as well

wiktor2200:
Manual grafana installation is not affected, because grafana service name is set here: https://github.com/Altinity/clickhouse-operator/blob/eda1163344acfe01dc0803c0a4de010a118920a5/deploy/grafana/grafana-manually/grafana.yaml#L69
So it's defined inside repo. 

`grafana-operator` constants are defined in outer repo which I mentioned in PR description.
What is more it's been already fixed here: 
* https://github.com/Altinity/clickhouse-operator/blob/5ecc53e1edf587cb82788e2e5922172cca2b1d34/deploy/grafana/grafana-with-grafana-operator/proxy-grafana.sh#L5
* https://github.com/Altinity/clickhouse-operator/blob/5ecc53e1edf587cb82788e2e5922172cca2b1d34/Vagrantfile#L104

wiktor2200:
Sure. I will add it to my form.

wiktor2200:
@Slach I've changed it. I've found one more occurrence here: `docs/k8s_cluster_access.md` so I've changed it as well.

PR #353, opened 2020-04-21
add env variable used to populate inter server http host

garrettlish:
Sometime FQDN is not reliable and rely on external DNS service reliability, we might need to expose IP used by other servers to access this sever.

Signed-off-by: xiancli <xiancli@ebay.com>

### DIFF
--- a/docs/operator_configuration.md
+++ b/docs/operator_configuration.md
@@ -98,6 +98,9 @@ chConfigUserDefaultNetworksIP:
   - "::/0"
 chConfigUserDefaultPassword: "default"
 
+## Default env variable used to populate inter server http host
+chConfigInterServerHttpHost: "pod_ip"
+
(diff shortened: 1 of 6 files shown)


garrettlish:
Sample config in `config.yaml` is as the following:
```
## Default env variable used to populate inter server http host
chConfigInterServerHttpHost: "pod_ip"
```

garrettlish:
@sunsingerus can you help review it?

garrettlish:
we can use `ChiConfiguration.Files` to achieve the same target, the change is meaningless

alex-zaitsev:
Thanks, @garrettlish. Yes, you can do it with files section. I've added an issue to have such a support on spec level as well: https://github.com/Altinity/clickhouse-operator/issues/354

PR #405, opened 2020-05-27
Improve prometheus alerts

Slach:
add 20 triggers


deploy/prometheus/prometheus.yaml
```diff
@@ -14,12 +14,16 @@ spec:
...
+      prometheus: prometheus
   enableAdminAPI: false
-#  alerting:
-#    alertmanagers:
-#    - namespace: default
-#      name: alertmanager
-#      port: web
+  alerting:
+    alertmanagers:
+    - namespace: ${PROMETHEUS_NAMESPACE}
```
sunsingerus, line 24: I would strictly suggest to rename `prometheus.yaml` into `prometheus-template.yaml`, since it has templated parameters from now

deploy/prometheus/create-prometheus.sh
```diff
@@ -53,7 +56,9 @@ kubectl --namespace="${PROMETHEUS_NAMESPACE}" apply --validate="${VALIDATE_YAML}
     https://raw.githubusercontent.com/coreos/prometheus-operator/${PROMETHEUS_OPERATOR_BRANCH}/example/rbac/prometheus/prometheus-service-account.yaml
 
 # Setup Prometheus instance via prometheus-operator into dedicated namespace
-kubectl --namespace="${PROMETHEUS_NAMESPACE}" apply --validate="${VALIDATE_YAML}" -f prometheus.yaml
+kubectl --namespace="${PROMETHEUS_NAMESPACE}" apply --validate="${VALIDATE_YAML}" -f <( \
+    cat prometheus.yaml | PROMETHEUS_NAMESPACE=${PROMETHEUS_NAMESPACE} envsubst \
```
sunsingerus, line 60: File naming

PR #508, opened 2020-08-20
Tests troubleshooting

sunsingerus:


tests/test_operator.py
```diff
@@ -373,7 +360,11 @@ def test_014():
         kubectl("delete pod zookeeper-0")
         kube_wait_object("pod", "zookeeper-0")
```
alex-zaitsev, line 361: Let's move it after the insert attempt.

alex-zaitsev (approved):

PR #542, opened 2020-09-17
tests: race conditions for volume tests

sunsingerus:


tests/test_operator.py
```diff
@@ -1040,7 +1040,12 @@ def test_021(config="configs/test-021-rescale-volume.yaml"):
             assert size == "50Mi"
 
         with And("There should be two disks recognized by ClickHouse"):
-            out = clickhouse.query(chi, "select count() from system.disks")
+            # ClickHouse requires some time to mount volume. Race conditions.
```
alex-zaitsev, line 1043: I don't like 120s timeout. Let's restart a pod for changes to be picked up quickly

PR #555, opened 2020-09-28
fix DiskFreeBytes and add DiskTotalBytes

Slach:
Signed-off-by: Eugene Klimov <eklimov@altinity.com>


alex-zaitsev (commented):
Looks ok, but I am concerned that it would not show a problem as Gennady's alert did

Slach:
It need to more complex refactoring for these metrics

Slach:
@alz please look to PR again

PR #565, opened 2020-10-14
Add username and password to clickhouse-client invocation

stumitchell:
Changed quickstart docs to add username and password to clickhouse-client invocation.

It seems that the default configuration doesn't allow the `default` user to log in from outside localhost.


Slach (approved):

Slach:
thanks a lot for your contribution

PR #598, opened 2020-11-23
fixes for pod_name into Zookeeper_dashboard.json

Slach:
- Zookeeper dashboard variables show only values for actual scraped pods
- Prometheus scraping config change, now we make pod scraping via annotations now instead of `kind:ServiceMonitor`
- update prometheus-operator to 0.43 version

Signed-off-by: Eugene Klimov <eklimov@altinity.com>


deploy/prometheus/prometheus-template.yaml
```diff
@@ -1,3 +1,63 @@
...
+        action: replace
+        target_label: __metrics_path__
+
+      - source_labels: [__meta_kubernetes_namespace]
+        action: replace
+        target_label: namespace
+
+      - source_labels: [__meta_kubernetes_pod_name]
+        action: replace
+        target_label: pod_name
```
Slach, line 48: in our specific configs, we use `pod_name` instead of `pod`

grafana-dashboard/Zookeeper_dashboard.json
```diff
@@ -9363,14 +9363,14 @@
         "allValue": null,
         "current": {},
         "datasource": "${DS_PROMETHEUS}",
-        "definition": "label_values({job=~\"$job\"}, namespace)",
+        "definition": "label_values(znode_count{job=~\"$job\"}, namespace)",
```
Slach, line 9366: now, variables values filters properly, all values related to Zookeeper

PR #358, opened 2020-04-24
Operator Helm Chart - draft

MatthiasLohr:
This is a first draft of a Helm Chart for deploying the operator on Kubernetes using Helm.

This pull request is opened according to the discussion in #262.

The complete configuration part is missing right now. Discussion in #262 might help to understand about possible configuration parameters and where to put into the files/ConfigMaps for the operator.

PR #619, opened 2020-12-16
delete an err code

TCeason:
The struct ChiReplica does not have port.

Signed-off-by: taichong <tai_chong@foxmail.com>


sunsingerus:
This is not the operator source code, just RnD. Should be deleted completely some time in future

TCeason:
> This is not the operator source code, just RnD. Should be deleted completely some time in future

When we use the End and what should I focus on if I want to contribute some code?

Slach:
@TCeason main parts of operator code contains in `cmd` and `pkg` directories,
I you want to help
you can try to help us with Helm chart
feel free continue work with https://github.com/Altinity/clickhouse-operator/pull/358

Issue #634, opened 2021-01-22
Duplicated user passwords when using password_double_sha1_hex


ben-efiz:
ClickHouse recommends using `password_double_sha1_hex` for compatibility with MySQL clients as decribed [here](https://clickhouse.tech/docs/en/operations/settings/settings-users/#user-namepassword).

When adding a `password_double_sha1_hex` as a user password, e.g. for admin, clickhouse-operator adds the `CHConfigUserDefaultPassword` as SHA256 encrypted password to the user. ClickHouse does not start if there is more then one password defined.

```
apiVersion: "clickhouse.altinity.com/v1"
kind: "ClickHouseInstallation"
metadata:
  name: "settings-01"
spec:
  configuration:
    users:
      # admin use has 'password_double_sha1_hex' so actual password value is not published
      admin/password_double_sha1_hex: b285a72aa0bf81129c828e8063fcd4a8f29b6a10
      admin/networks/ip: "127.0.0.1/32"
      admin/profile: default
      admin/quota: default
    clusters:
      - name: "standard"
        layout:
          shardsCount: 1
          replicasCount: 1
```

sunsingerus:
PR accepted, this issue should be closed as soon as test cycle for 0.13.5 completed

sunsingerus:
done

PR #635, opened 2021-01-22
#634 Checking if also password_double_sha1_hex is present 

ben-efiz:
This is just a tiny fix. I understand it needs testing and an example yaml. Please let me know which of the many possible places to add it to,


sunsingerus (approved):

sunsingerus:
Thanks a lot!

PR #667, opened 2021-03-06
Optimization of pod batch restart in cluster upgrade

xiedeyantu:
Optimize the statefulset update, patching labels first will not trigger forbidden error when updating statefulset.
codes will not enter the subsequent process of deleting and reconstructing statefulset.
Pods are no longer restarted when new shards/replicas are added, modified and deleted.

### DIFF
--- a/pkg/controller/chi/worker.go
+++ b/pkg/controller/chi/worker.go
@@ -253,7 +253,7 @@ func (w *worker) updateCHI(old, new *chop.ClickHouseInstallation) error {
 		WithStatusAction(new).
 		M(new).F().
 		Info("reconcile started")
-	w.a.V(2).M(new).F().Info("action plan\n%s\n", actionPlan.String())
+	//w.a.V(2).M(new).F().Info("action plan\n%s\n", actionPlan.String())
 
 	if new.IsStopped() {
 		w.a.V(1).
--- a/pkg/controller/chi/creator.go
+++ b/pkg/controller/chi/creator.go
@@ -50,12 +50,18 @@ func (c *Controller) createStatefulSet(statefulSet *apps.StatefulSet, host *chop
 // updateStatefulSet is an internal function, used in reconcileStatefulSet only
 func (c *Controller) updateStatefulSet(oldStatefulSet *apps.StatefulSet, newStatefulSet *apps.StatefulSet, host *chop.ChiHost) error {
 	log.V(2).M(host).F().P()
-	// update labels
-	diffLabelsBytes := DiffLabels(oldStatefulSet, newStatefulSet)
+
+	// diff labels
+	diffLabelsBytes, err := DiffLabels(oldStatefulSet, newStatefulSet)
+	if err != nil {
+		log.V(1).M(host).A().Error("%v", err)
+		return err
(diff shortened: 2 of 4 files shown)


xiedeyantu:
@Slach

xiedeyantu:
@sunsingerus

xiedeyantu:
@alex-zaitsev could you take a look at this pr please?

xiedeyantu:
@sunsingerus Excuse me, is there any problem with this pr, or is there any better solution to this problem?

pkg/controller/chi/worker.go
```diff
@@ -253,7 +253,7 @@ func (w *worker) updateCHI(old, new *chop.ClickHouseInstallation) error {
 		WithStatusAction(new).
 		M(new).F().
 		Info("reconcile started")
-	w.a.V(2).M(new).F().Info("action plan\n%s\n", actionPlan.String())
+	//w.a.V(2).M(new).F().Info("action plan\n%s\n", actionPlan.String())
```
aksdb, line 256: Was it intentional to disable this log line? (If so: why not remove it completely?)

xiedeyantu: When debugging, this will report null pointer exception. I wonder if it is an environment problem, so do this.

aksdb:
I tried this patchset / PR in our cluster, since I had a similar problems (pods being restarted completely when labels change), but now they don't update at all. If I change the clickhouse version in a podtemplate and apply the new custom resource definition, no new changes are rolled out. The labels are updated, but that's it :-/
Without this PR those changes are immediately applied.

alex-zaitsev:
> I tried this patchset / PR in our cluster, since I had a similar problems (pods being restarted completely when labels change), but now they don't update at all. If I change the clickhouse version in a podtemplate and apply the new custom resource definition, no new changes are rolled out. The labels are updated, but that's it :-/
> Without this PR those changes are immediately applied.

PR is incorrect. In order to solve the original issue we will remove those confusing labels at all. They were more for debugging purposes.

xiedeyantu:
> I tried this patchset / PR in our cluster, since I had a similar problems (pods being restarted completely when labels change), but now they don't update at all. If I change the clickhouse version in a podtemplate and apply the new custom resource definition, no new changes are rolled out. The labels are updated, but that's it :-/
> Without this PR those changes are immediately applied.

This PR is to solve the problem of pod restart. The pod tag has forgotten to update in the code. You can be updated without my PR, which will inevitably lead to the restart of pod.

xiedeyantu:
> > I tried this patchset / PR in our cluster, since I had a similar problems (pods being restarted completely when labels change), but now they don't update at all. If I change the clickhouse version in a podtemplate and apply the new custom resource definition, no new changes are rolled out. The labels are updated, but that's it :-/
> > Without this PR those changes are immediately applied.
> 
> PR is incorrect. In order to solve the original issue we will remove those confusing labels at all. They were more for debugging purposes.

Now, each sts only controls one pod. I don't think there is a need to configure tags on pods. Just configure tags on sts. I don't know what's wrong with my PR.

Issue #487, opened 2020-08-03
Need a way to limit what labels get copied to child resources.


denniswebb:
## Issue and Background

We use [ArgoCD](https://github.com/argoproj/argo-cd) for our GitOps operator. Our ClickHouseInstallation is `kubectl apply` itself with Argo. If we remove the YAML for an object from our git repo, the ArgoCD will see this has been removed and will remove the old resource.

ArgoCD is able to handle this buy adding a label `argocd.argoproj.io/instance` to all Kubernetes resources it has created so it knows if the YAML for that resource is removed from github, that it was originally managed by ArgoCD and should be pruned(deleted).

The issue with ClickhouseOperator is that it copies all the labels from the ClickHouseInstallation resource and adds it to the child resources such as configmaps, services, pods, etc. This also copies the `argocd.argoproj.io/instance` and triggers ArgoCD to believe that the resource has been abandoned and it wants to prune those configmaps, statefulsets, etc.

This can be avoided if an object has the annotation `argocd.argoproj.io/sync-options: Prune=false` attached to it, but unfortunately ClickHouseOperator does not copy the annotations to all resources.

## Proposed Solution

I believe the best solution to this would be to add a configuration option to any ClickHouseInstallation for a list of `IgnoreLabels`. We could configure any non-clickhouse related labels that shouldn't be copied here. This will prevent the `argocd.argoproj.io/instance` label or any other label we choose from being copied to all resources managed and owned by the ClickHouseOperator process.

alex-zaitsev:
@denniswebb , this is an interesting use case. Maybe it is easier to add a flag that would prevent copying user labels from CHI to dependent resources at all?

denniswebb:
That works for me.

sunsingerus:
Implemented in #676 
PR merged. 
To be released in 0.14.0

PR #676, opened 2021-04-12
Support label exclusions

aksdb:
As mentioned in issue #487, CD systems like Flux or Argo tend to put labels on resources they manage. If those labels get propagated, weird stuff can start happening.

This PR introduces a new config settings `labelsIgnore` which will filter those labels before propagating them.

### DIFF
--- a/pkg/model/labeler.go
+++ b/pkg/model/labeler.go
@@ -297,7 +297,8 @@ func (l *Labeler) GetSelectorHostScope(host *chi.ChiHost) map[string]string {
 
 // appendCHILabels appends CHI-provided labels to labels set
 func (l *Labeler) appendCHILabels(dst map[string]string) map[string]string {
-	return util.MergeStringMapsOverwrite(dst, l.chi.Labels)
+	sourceLabels := util.CopyMapExcept(l.chi.Labels, l.chop.Config().LabelsIgnore...)
+	return util.MergeStringMapsOverwrite(dst, sourceLabels)
 }
 
 // appendReadyLabels appends "Ready" label to labels set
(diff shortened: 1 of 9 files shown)


sunsingerus (changes requested):
Please, prepare PR as specified in [this doc](https://github.com/Altinity/clickhouse-operator/blob/master/CONTRIBUTING.md#how-to-make-pr)
Next version to be released - 0.14.0

pkg/model/labeler.go
```diff
@@ -277,7 +277,13 @@ func (l *Labeler) GetSelectorHostScope(host *chi.ChiHost) map[string]string {
 
 // appendCHILabels appends CHI-provided labels to labels set
 func (l *Labeler) appendCHILabels(dst map[string]string) map[string]string {
-	return util.MergeStringMapsOverwrite(dst, l.chi.Labels)
+	sourceLabels := l.chi.Labels
+	if len(l.chop.Config().LabelsIgnore) > 0 {
+		sourceLabels = util.MergeStringMapsOverwrite(map[string]string{}, sourceLabels)
```
sunsingerus, line 282: Probably, it would be clearer to introduce new util function to make a Copy of a map.
Merging over an empty map is not that obvious

aksdb: Good point, that makes for cleaner code. I added two helper functions. (`CopyMap` and `CopyMapExcept`)

aksdb:
> Please, prepare PR as specified in [this doc](https://github.com/Altinity/clickhouse-operator/blob/master/CONTRIBUTING.md#how-to-make-pr)
> Next version to be released - 0.14.0

:facepalm: I had *one* job .... sorry, for completely missing those two completely simple points.

sunsingerus (approved):

sunsingerus:
Thanks a lot!

PR #680, opened 2021-04-22
Add detached parts monitoring

Slach:


pkg/apis/metrics/clickhouse_fetcher.go
```diff
@@ -121,6 +121,21 @@ const (
 			toString(total_space) AS total_space			
         FROM system.disks
 	`
+
+	queryDetachedPartsSQL = `
+		SELECT 
+			count() AS detached_parts,
+			database,
+			table,
+			disk,
```
alex-zaitsev, line 130: I do not think disk is important here.

Slach: I think it will important if you will have to detached part cause only one disk will have bitflip \ bios error etc.
I agree, usually `disk` label will have a `default` value

Issue #686, opened 2021-04-27
How to install a ClickHouse Cluster on other ns?


TCeason:
## Summary

| ClickHouse-Operator Version |
|:--:|
| 0.14.0 |

If I install a CHOP on ns `foo`, and I want to install the ClickHouse cluster in other ns `boo`, how to do it?

## Scene

I install `chop`  on ns `chop-altinity` and then I try to install a [simple.yaml](https://github.com/Altinity/clickhouse-operator/blob/master/docs/chi-examples/01-simple-layout-01-1shard-1repl.yaml) in ns `default`.

```
$ kubectl get all  -n chop-altinity 
NAME                                      READY   STATUS    RESTARTS   AGE
pod/clickhouse-operator-5c4cf66cc-wvsbz   2/2     Running   0          50m

NAME                                  TYPE        CLUSTER-IP    EXTERNAL-IP   PORT(S)    AGE
service/clickhouse-operator-metrics   ClusterIP   10.96.36.23   <none>        8888/TCP   50m

NAME                                  READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/clickhouse-operator   1/1     1            1           50m

NAME                                            DESIRED   CURRENT   READY   AGE
replicaset.apps/clickhouse-operator-5c4cf66cc   1         1         1       50m

$ kubectl exec -it clickhouse-operator-5c4cf66cc-wvsbz    -n chop-altinity -c clickhouse-operator -- /clickhouse-operator --version 
0.14.0
```

```
$ kubectl get chi              
NAME        CLUSTERS   HOSTS   STATUS
simple-01
```

@Slach

Slach:
you shall change kubernetes configmap `etc-clickhouse-operator-files` in namespace `test-chop`
and add to data->`config.yaml` key    

```
watchNamespaces:
  - test-cluster
  - test-chop
```

TCeason:
[...]

Issue #294, opened 2020-03-09
Make it possible to watch all namespaces from non kube-system namespace


wilmardo:
Currently deploying the operator in another namespace then `kube-system` implicitly disallows global watching:
https://github.com/Altinity/clickhouse-operator/blob/aaee541b6d2b1ae79aa2d11088ec9c7f377a1dc3/pkg/apis/clickhouse.altinity.com/v1/type_config.go#L323-L329

It would be nice to have a parameter for this behavior, we deploy all operators in the `operator` namespace and they are cluster global.

I am open to implement this but what is a nice approach to this? A `WATCH_ALL_NAMESPACES` boolean? When `WATCH_NAMESPACES` or `WATCH_NAMESPACE` is defined it will be ignored.

PS why not consolidate `WATCH_NAMESPACE` and `WATCH_NAMESPACES` into just `WATCH_NAMESPACES`? Nothing wrong with a list with one entry or am I missing something? Feels weird as enduser :)

anxolerd:
I'd also like to have operator installed in its own namespace, but watching all namespaces.

By the way, I've heard, it is common practice for users of operator-framework to watch all namespaces if `WATCH_NAMESPACE` environment variable is empty.

TCeason:
> I'd also like to have operator installed in its own namespace, but watching all namespaces.
> 
> By the way, I've heard, it is common practice for users of operator-framework to watch all namespaces if `WATCH_NAMESPACE` environment variable is empty.

If set empty just only watch the ns where operator was installed.

https://github.com/Altinity/clickhouse-operator/issues/686

Slach:
sorry for late response, solution here
https://github.com/Altinity/clickhouse-operator/issues/1418#issuecomment-2111986205

PR #695, opened 2021-05-07
Default watch all Namespaces

TCeason:
If config.WatchNamespaces is empty we should watch all Namseapces in the k8s.
https://github.com/Altinity/clickhouse-operator/issues/686

### DIFF
--- a/pkg/apis/clickhouse.altinity.com/v1/type_config_chop.go
+++ b/pkg/apis/clickhouse.altinity.com/v1/type_config_chop.go
@@ -474,14 +474,8 @@ func (config *OperatorConfig) applyDefaultWatchNamespace() {
 	}
 
 	// No namespaces specified
-
-	namespace := os.Getenv(OPERATOR_POD_NAMESPACE)
-	if namespace == "kube-system" {
-		// Do nothing, we already have len(config.WatchNamespaces) == 0
-	} else {
-		// We have WATCH_NAMESPACE specified
-		config.WatchNamespaces = []string{namespace}
-	}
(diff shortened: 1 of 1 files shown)


TCeason:
@Slach  Please review this pr and present your opinion. Thanks.

Slach:
look like it will break behavior when you install separate clickhouse-operator to separate "tenant" namespace for multi-tenant kubernetes cluster like OpenShift

TCeason:
If so, how to resolve this question: https://github.com/Altinity/clickhouse-operator/issues/294?

TCeason:
The value of unspecified watchNamespaces should be the same as the value of specified kube-system.

Because :  In case no namespaces specified - watch all namespaces

```
// TODO unify with GetInformerNamespace
// IsWatchedNamespace returns whether specified namespace is in a list of watched
func (config *OperatorConfig) IsWatchedNamespace(namespace string) bool {
	// In case no namespaces specified - watch all namespaces
	if len(config.WatchNamespaces) == 0 {
		return true
	}

	return util.InArray(namespace, config.WatchNamespaces)
}

```

And befort this this pr only `namespace == "kube-system"` the len(config.WatchNamespaces) equal with 0.

Slach:
@sunsingerus @alex-zaitsev could share your opinion?

TCeason:
Ok, let me see

Slach:
@TCeason great ;) thanks a lot for your efforts!

Slach (changes requested):
propose add regexp or glob support for watchNamespaces

PR #718, opened 2021-06-09
modify grafana operator install sh

TCeason:
The grafana fold has already changed.

https://github.com/integr8ly/grafana-operator/tree/master/deploy/olm-catalog/grafana-operator/3.x.x


Slach:
already fixed on 0.15.0
also 3.10.x operator doesn't look stable now

PR #725, opened 2021-06-20
Update clickhouse-operator-install.sh

nrmitchi:
Documentation states the default installation namespace is `clickhouse-operator`, not `kube-system`.

It is also `clickhouse-operator` in the deletion script, so this just makes them match.

I think this counts as "supplementary material", so PR is against master. Feel free to change if necessary.

### DIFF
--- a/deploy/operator-web-installer/clickhouse-operator-install.sh
+++ b/deploy/operator-web-installer/clickhouse-operator-install.sh
@@ -81,7 +81,7 @@ function get_file() {
 #
 
 # Namespace to install operator
-OPERATOR_NAMESPACE="${OPERATOR_NAMESPACE:-kube-system}"
+OPERATOR_NAMESPACE="${OPERATOR_NAMESPACE:-clickhouse-operator}"
 METRICS_EXPORTER_NAMESPACE="${OPERATOR_NAMESPACE}"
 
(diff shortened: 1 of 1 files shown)


Slach:
https://github.com/Altinity/clickhouse-operator/blob/master/docs/quick_start.md#clickhouse-operator-installation
describe following things
- you need to define OPERATOR_NAMESPACE variable before run installation script, to install clickhouse-operator to selected namespace explicitly
- also, when clickhouse-operator install in the separate namespace, it watches only this namespace, but if you install clickhouse-operator into `kube-system` it will watch custom resources like a `kind: ClickhouseInstallation` in all namespaces

nrmitchi:
That's fine, but if this is the case then the instal/delete script should _both_ use `kube-system` as the namespace, and the [documentation](https://github.com/Altinity/clickhouse-operator/blob/master/docs/quick_start.md#in-case-youd-like-to-customize-installation-parameters) should not state:

> If no OPERATOR_NAMESPACE specified,... installer will create namespace clickhouse-operator and install clickhouse-operator into it.

Slach:
@nrmitchi thanks a lot for clarifying and your efforts for contribution, i hope latest documentation changes will help community

PR #736, opened 2021-07-08
separate `ClickHouseMetricsExporterFetchErrors` from `ClickHouseServerDown` and remove "fail fast" strategy from metrics-exporter

Slach:
After discussion with @alex-zaitsev we decide a little bit  improve alerts and metrics-exporter

Signed-off-by: Slach <bloodjazman@gmail.com>

### DIFF
--- a/deploy/prometheus/prometheus-alert-rules-clickhouse.yaml
+++ b/deploy/prometheus/prometheus-alert-rules-clickhouse.yaml
@@ -22,7 +22,7 @@ spec:
               Please check instance status
               ```kubectl logs -n {{ $labels.namespace }} {{ $labels.pod_name }} -c metrics-exporter -f```
 
-        - alert: ClickHouseServerDown
+        - alert: ClickHouseServerPossibleDown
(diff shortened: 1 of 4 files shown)


deploy/prometheus/prometheus-alert-rules-clickhouse.yaml
```diff
@@ -22,7 +22,7 @@ spec:
               Please check instance status
               ```kubectl logs -n {{ $labels.namespace }} {{ $labels.pod_name }} -c metrics-exporter -f```
 
-        - alert: ClickHouseServerDown
+        - alert: ClickHouseServerPossibleDown
           expr: chi_clickhouse_metric_fetch_errors > 0
```
alex-zaitsev, line 26: Can we have the critical alert only system.metrics fetch error? 
Others may happen at running ClickHouse server, and I would like to see an alert but not a down one.

Slach: yep, I can split this alert for two
critical for fetch_type='system_metrics' and high for all other types

PR #815, opened 2021-11-05
Add spec section to 1 shard 1 replica example

ghjm:
Signed-off-by: Graham Mainwaring <gmainwaring@altinity.com>

When trying to follow the quick start instructions with minikube 1.22, I get the following error:

```
$ kubectl apply -n default -f https://raw.githubusercontent.com/Altinity/clickhouse-operator/master/docs/chi-examples/01-simple-layout-01-1shard-1repl.yaml
error: error validating "https://raw.githubusercontent.com/Altinity/clickhouse-operator/master/docs/chi-examples/01-simple-layout-01-1shard-1repl.yaml": error validating data: ValidationError(ClickHouseInstallation): missing required field "spec" in com.altinity.clickhouse.v1.ClickHouseInstallation; if you choose to ignore these errors, turn validation off with --validate=false
```

Adding an explicit `spec:` section makes it work for me.  I speculate that at some time in the past, the operator allowed the `spec:` section to be missing, and defaulted to a 1 shard, 1 replica configuration, but now it is required.


alex-zaitsev (changes requested):
Thanks for the catch! 'spec' is required, but cluster layout is not. The simplest cluster will look like:

```
spec:
  configuration:
    clusters:
      - name: single
 ```

Slach (approved):

Slach (approved):

PR #870, opened 2022-01-27
fix non pointer mistake in metrics-exporter

adu-bricks:
Signed-off-by: Anthony Dubuissez <a.dubuissez@bricks-co.com>

Fixing the metrics-exporter report of the system.parts fetch that would never return any data due to a missing pointer.

I put the master branch for this pr as I didnt find the next-release branch, only 0.18.0.

* [x ] All commits in the PR are squashed. [More info](https://github.com/Altinity/clickhouse-operator/blob/master/CONTRIBUTING.md#how-to-make-pr) 
* [ x] The PR is made into dedicated `next-release` branch, **not into** `master` branch<sup>1</sup>. [More info](https://github.com/Altinity/clickhouse-operator/blob/master/CONTRIBUTING.md#how-to-make-pr)
* [ x] The PR is signed. [More info](https://github.com/Altinity/clickhouse-operator/blob/master/CONTRIBUTING.md#sign-your-work)

### DIFF
--- a/pkg/apis/metrics/clickhouse_fetcher.go
+++ b/pkg/apis/metrics/clickhouse_fetcher.go
@@ -171,7 +171,7 @@ func (f *ClickHouseFetcher) getClickHouseSystemParts() ([][]string, error) {
 				metricDiskDataBytes, metricMemoryPrimaryKeyBytesAllocated string
 			if err := rows.Scan(
 				&database, &table, &active, &partitions, &parts, &bytes, &uncompressed, &_rows,
-				&metricDiskDataBytes, metricMemoryPrimaryKeyBytesAllocated,
+				&metricDiskDataBytes, &metricMemoryPrimaryKeyBytesAllocated,
 			); err == nil {
 				*data = append(*data, []string{
 					database, table, active, partitions, parts, bytes, uncompressed, _rows,


Slach:
Thanks a lot for contribution, I wonder how it works before without errors

adu-bricks:
> Thanks a lot for contribution, I wonder how it works before without errors

it does error (albeit silently), there is only a check `if err == nill` and so if there is one the err is never pushed to the upstream func to report it.

Slach:
> it does error (albeit silently), there is only a check if err == nill and so if there is one the err is never pushed to the upstream func to report it.

but, this numbers is present in our Grafana Dashboard and show 24x7, correcly now

adu-bricks:
> > it does error (albeit silently), there is only a check if err == nill and so if there is one the err is never pushed to the upstream func to report it.
> 
> but, this numbers is present in our Grafana Dashboard and show 24x7, correcly now

after some investigation, it. look like the new code never worked, this metrics as been broken since 0.17.0, last working docker image for the metrics-exporter that work is the 0.16.1.

https://github.com/Altinity/clickhouse-operator/commit/28af2b0b6e9ca322a4260601a5f2dbb0a7ac97ae is the commit when it started to never get a value other than 0 

Looking at the history it look like per default the image for the metrics-exporter is shipped with :latest but this tag was not in sync with the master branch before ?

Slach:
@adu-bricks thanks a lot for detailed investigation!

Slach:
@adu-bricks could you check and confirm issue fixed on metrics-exporter  0.18.0 and latest image version?

adu-bricks:
> @adu-bricks could you check and confirm issue fixed on metrics-exporter 0.18.0 and latest image version?

I can confirm pulling the news :latest and :0.18.0 image fixed the metrics issues.

Thanks !

PR #872, opened 2022-01-28
Fix GitHub workflows

Slach:


.github/workflows/run_tests.yaml
```diff
@@ -6,7 +6,25 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
+      - run: |
+          sudo apt-get update
+          sudo apt-get install -y conntrack
+      - uses: medyagh/setup-minikube@latest
+        with:
+          driver: docker
+          container-runtime: containerd
+          kubernetes-version: v1.21.9
```
zvonand, line 16: why not 1.23?

tests/image/Dockerfile
```diff
@@ -79,7 +79,7 @@ EXPOSE 2375
 
 ENV MINIKUBE_HOME /home/master/.minikube
 
-ENV K8S_VERSION=${K8S_VERSION:-1.22.2}
+ENV K8S_VERSION=${K8S_VERSION:-1.21.6}
```
zvonand, line 82: Maybe we should use the same k8s version all way long to keep things simpler?

.github/workflows/run_tests.yaml
Slach: potentially current prometheus-operator and grafana-operator not works on 1.22+, we need upgrade it first in separate PR it need research, for example grafana-operator 4.x version broke installation process compatible with 3.9.x

tests/image/Dockerfile
Slach: 1,21,6 it's last 1.21 version available on minikube preloaded archives

zvonand (approved):
👍🏻

PR #883, opened 2022-02-10
Add clickhouse keeper

Slach:


Slach:
@alex-zaitsev ready for review, let's do it?

deploy/clickhouse-keeper/clickhouse-keeper-1-node-256M-for-test-only.yaml
```diff
@@ -0,0 +1,246 @@
...
+            <level>trace</level>
+            <console>true</console>
+        </logger>
+        <listen_host>0.0.0.0</listen_host>
+        <keeper_server incl="keeper_server">
+            <path>/var/lib/clickhouse-keeper</path>
+            <tcp_port>2181</tcp_port>
+            <coordination_settings>
+                <!-- <raft_logs_level>trace</raft_logs_level> -->
+                <raft_logs_level>debug</raft_logs_level>
```
filimonov, line 85: check also https://github.com/ClickHouse/ClickHouse/blob/bc2d0ee7c7c9be67825ae3f8acf8a9f08bc7c331/tests/config/config.d/keeper_port.xml#L17

deploy/clickhouse-keeper/clickhouse-keeper-1-node-256M-for-test-only.yaml
```diff
@@ -0,0 +1,246 @@
...
+  labels:
+    app: clickhouse-keeper
+spec:
+  selector:
+    matchLabels:
+      app: clickhouse-keeper
+  serviceName: clickhouse-keepers
+  replicas: 1
+  updateStrategy:
+    type: RollingUpdate
```
filimonov, line 106: Actually to reduce the number of leader shifts the best strategy is restarting the leader node last (same as for zookeeper). But i don't know if it can be programmed somehow in k8. (simple option should also work)

deploy/clickhouse-keeper/clickhouse-keeper-1-node-256M-for-test-only.yaml
```diff
@@ -0,0 +1,246 @@
...
+                echo "<yandex><keeper_server>"
+                echo "<server_id>${MY_ID}</server_id>"
+                echo "<raft_configuration>"
+                for (( i=1; i<=$SERVERS; i++ )); do
+                    echo "<server><id>${i}</id><hostname>$NAME-$((i-1)).${DOMAIN}</hostname><port>${RAFT_PORT}</port></server>"
+                done
+                echo "</raft_configuration>"
+                echo "</keeper_server></yandex>"
+              } >> /tmp/clickhouse-keeper/config.d/generated-keeper-settings.xml &&
+              cat /tmp/clickhouse-keeper/config.d/generated-keeper-settings.xml &&
```
filimonov, line 182: btw - why /tmp?

deploy/clickhouse-keeper/clickhouse-keeper-1-node-256M-for-test-only.yaml
```diff
@@ -0,0 +1,246 @@
...
+            - "-listen"
+            - 0.0.0.0:7000
+            - "-zk-hosts"
+            - 127.0.0.1:2181
+          livenessProbe:
+            exec:
+              command:
+                - sh
+                - -c
+                - "OK=$(echo ruok | nc 127.0.0.1 2181); if [[ \"$OK\" == \"imok\" ]]; then exit 0; else exit 1; fi"
```
filimonov, line 234: that exporter - may be just check that 7000 is opened?

Slach: No! This is the proper place of liveness probe. 
It allows us to properly monitor `up` metrics for /metrtics endpoint from prometehus side, when keeper down then exporter also shall down, otherwise prometheus will continue scraping exporter and will always get metrics (look like it's  a just specified zabbix-exporter behavior)

deploy/clickhouse-keeper/clickhouse-keeper-1-node-256M-for-test-only.yaml
Slach: Cause double mount ConfigMap + temporary generated config in the same folder `/etc/clickhouse-keeper` + `/etc/clickhouse-keeper/config.d/` not possible in Kubernetes

deploy/clickhouse-keeper/clickhouse-keeper-1-node-256M-for-test-only.yaml
filimonov: OK. But how to detect the situation when the keeper is up, but the exporter is down / can't start?

deploy/clickhouse-keeper/clickhouse-keeper-1-node-256M-for-test-only.yaml
filimonov: I think we can mount single file /etc/clickhouse-keeper/keeper_config.xml as a volume. 

Storing configs in /tmp is not a good practice (due to security reasons).

deploy/clickhouse-keeper/clickhouse-keeper-1-node-256M-for-test-only.yaml
Slach: We already will catch this corner case, when `keeper` is up, but `zookeer-exporter` is stuck, we will have `up == 0` on `prometheus` side because this metric calculates on `prometheus` side not inside `zookeeper-exporter`

deploy/clickhouse-keeper/clickhouse-keeper-1-node-256M-for-test-only.yaml
Slach: I checked, for `raft_logs_level debug` I see only changes in quorum + ruok/mntr execution command from liveness / readyness probes, for trace log 

I don't think we need to change anything else on this ConfigMap related with standalone mode execution

deploy/clickhouse-keeper/clickhouse-keeper-1-node-256M-for-test-only.yaml
Slach: We have `kind: PodDisruptionBudget` to avoid situation when more than one pod will restart on normal cases

deploy/clickhouse-keeper/clickhouse-keeper-1-node-256M-for-test-only.yaml
Slach: I tried, even when I mount only one key from ConfigMap with read/write mode, I can't dynamically write into config.d/ subfolder

/tmp/ is ok in this case, cause we create this config on every restart, before starting `keeper` daemon itself

deploy/clickhouse-keeper/clickhouse-keeper-1-node-256M-for-test-only.yaml
filimonov: But from k8 side it will look as 'everything is ok'. OK, I hope that exporter will not stuck often :)

Issue #859, opened 2022-01-11
0.17.0 Config excludeFromPropagationAnnotations not excluding annotations for various created resources


tanenbaum:
If I set up a CHOP with config.yaml containing:

```
excludeFromPropagationAnnotations: ["foo"]
```
and a CHI with metadata like:

```
metadata:
  annotations:
    foo: bar
```
I would expect all created resources to be without the "foo" annotation. 

I actually find that the StatefulSets, ConfigMaps, PVCs and host specific Services have the annotation.

Tracing through the code, I think https://github.com/Altinity/clickhouse-operator/blob/dc6cdc6f2f61fc333248bb78a8f8efe792d14ca2/pkg/model/annotator.go#L112 is where the problem lies. 

```
func (a *Annotator) getHostScope(host *chiv1.ChiHost) map[string]string {
	return a.appendCHIProvidedTo(host.GetAnnotations())
}
```
Any function using `getHostScope` ends up with all the CHI annotations as `a.appendCHIProvidedTo(host.GetAnnotations())` is passing in the CHI annotations that take precedence over any of the exclusions. Other similar functions here use `a.appendCHIProvidedTo(nil)` instead.

Perhaps `(a *Annotator) appendCHIProvidedTo(dst map[string]string)` should not take an argument here, as there are no defaults used as there are with labels and therefore should always be `nil`.

sunsingerus:
This issue was addressed in 0.18.2 version, please check

tanenbaum:
I will be testing this change soon (next day or so). I'll get back to you
asap.

alex-zaitsev:
Fixed in  https://github.com/Altinity/clickhouse-operator/releases/tag/0.18.2

PR #884, opened 2022-02-11
Added annotations in metadata propagation test

zvonand:
Added new test covering #859

### DIFF
--- a/tests/e2e/test_operator.py
+++ b/tests/e2e/test_operator.py
@@ -1983,9 +1983,9 @@ def test_031(self):
             config_yaml["annotation"]["exclude"] = ["excl", ]
             config_contents = yaml.dump(config_yaml, default_flow_style=False)
 
-            for i in manifest_yaml:
-                if i["metadata"]["name"] == "etc-clickhouse-operator-files":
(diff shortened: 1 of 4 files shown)


tests/e2e/test_operator.py
```diff
@@ -1967,6 +1969,58 @@ def test_030(self):
...
+    with Given("I generate CHO deploy manifest"):
+        with open(util.get_full_path(settings.clickhouse_operator_install_manifest)) as base_template,\
+                open(util.get_full_path("../../config/config.yaml")) as config_file:
+            manifest_yaml = list(yaml.safe_load_all(base_template.read()))
+
+            config_yaml = yaml.safe_load(config_file.read())
+            config_yaml["annotation"]["exclude"] = ["excl", ]
+            config_contents = yaml.dump(config_yaml, default_flow_style=False)
+
+            for i in manifest_yaml:
```
Slach, line 1986: i think "doc" instead of "i" will more mental here

Slach:
@alex-zaitsev 
Hm, why test is not correct? Test checks annotations excluding \ including propagation and change config for that, what is wrong in this scenario?

PR #895, opened 2022-03-02
Fix typo in Readme for Copyright

fuziontech:
Really basic typo fix. Could be mistaken though. Maybe you've managed to actually copyright all the way to 2219 ;)


Slach:
thanks a lot for your contibution, but is not the typo

PR #938, opened 2022-05-19
Add support for setting secure communication between clickhouse instances

### LINKED ISSUE #668: Secure remote_servers section
In order to use secure tcp port(9440 by default) while running distributed queries, it's need to add `secure` setting in `remote_servers` settings.
It would be great if clickhouse-operator could do that automatically (or at least we would be able to enable it via yaml) in case of using secure ports.

```
        <shard>
            <weight>1</weight>
            <internal_replication>true</internal_replication>
            <replica>
                <host>example01-02-1</host>
                <secure>1</secure>
                <port>9440</port>
            </replica>
            <replica>
                <host>example01-02-2</host>
                <secure>1</secure>
                <port>9440</port>
            </replica>
       <shard>
```

Also it would be great to have support for inter-server cluster secret.
https://github.com/ClickHouse/ClickHouse/pull/13156

chancez:

Fixes #668 

Verified this works when configuring clickhouse to use TLS and with no unencrypted ports.

A snippet of my CHI:
```
  templates:
    hostTemplates:
    - name: host-template
      spec:
        {{- if .Values.clickhouse.tls.enabled }}
        tcpPort: 9440
        secure: true
        httpPort: 8443
        interserverHTTPPort: 9010
        {{- else }}
        tcpPort: 9000
        httpPort: 8123
        interserverHTTPPort: 9009
        {{- end }}
```

### DIFF
--- a/deploy/builder/templates-install-bundle/clickhouse-operator-install-yaml-template-01-section-crd-01-chi-chit.yaml
+++ b/deploy/builder/templates-install-bundle/clickhouse-operator-install-yaml-template-01-section-crd-01-chi-chit.yaml
@@ -837,6 +837,11 @@ spec:
                                               allows connect to `clickhouse-server` via TCP Native protocol via kubernetes `Service`
                                             minimum: 1
                                             maximum: 65535
+                                          secure:
+                                            type: boolean
+                                            description: |
+                                              optional, setup `secure` inside `clickhouse-server` settings for each Pod where current template will apply
+                                              if specified
                                           httpPort:
                                             type: integer
                                             description: |
@@ -989,6 +994,11 @@ spec:
                                               allows connect to `clickhouse-server` via TCP Native protocol via kubernetes `Service`
                                             minimum: 1
                                             maximum: 65535
+                                          secure:
+                                            type: boolean
+                                            description: |
+                                              optional, setup `secure` inside `clickhouse-server` settings for each Pod where current template will apply
+                                              if specified
                                           httpPort:
                                             type: integer
                                             description: |
@@ -1105,6 +1115,11 @@ spec:
                                   More info: https://clickhouse.tech/docs/en/interfaces/tcp/
                                 minimum: 1
                                 maximum: 65535
+                              secure:
+                                type: boolean
+                                description: |
+                                  optional, setup `secure` inside `clickhouse-server` settings for each Pod where current template will apply
+                                  if specified
                               httpPort:
                                 type: integer
                                 description: |
--- a/deploy/operator/clickhouse-operator-install-ansible.yaml
+++ b/deploy/operator/clickhouse-operator-install-ansible.yaml
@@ -844,6 +844,11 @@ spec:
                                               allows connect to `clickhouse-server` via TCP Native protocol via kubernetes `Service`
                                             minimum: 1
                                             maximum: 65535
+                                          secure:
+                                            type: boolean
+                                            description: |
+                                              optional, setup `secure` inside `clickhouse-server` settings for each Pod where current template will apply
+                                              if specified
                                           httpPort:
                                             type: integer
                                             description: |
@@ -996,6 +1001,11 @@ spec:
                                               allows connect to `clickhouse-server` via TCP Native protocol via kubernetes `Service`
                                             minimum: 1
                                             maximum: 65535
+                                          secure:
+                                            type: boolean
+                                            description: |
+                                              optional, setup `secure` inside `clickhouse-server` settings for each Pod where current template will apply
+                                              if specified
                                           httpPort:
                                             type: integer
                                             description: |
@@ -1112,6 +1122,11 @@ spec:
                                   More info: https://clickhouse.tech/docs/en/interfaces/tcp/
                                 minimum: 1
                                 maximum: 65535
+                              secure:
+                                type: boolean
+                                description: |
+                                  optional, setup `secure` inside `clickhouse-server` settings for each Pod where current template will apply
+                                  if specified
                               httpPort:
                                 type: integer
                                 description: |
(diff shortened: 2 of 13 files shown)


chancez:
Looks like WalkHosts doesn't handle `hostTemplate`, so I need to rethink this a bit

alex-zaitsev (changes requested):
It would be ideal to add a test for that, but even an example manifest that shows the use of this feature would help.

Maybe some extra automation may be useful as well, e.g. automatically change default ports to secure when secure flag is used.

chancez:
Here's an example based on what we're using to deploy clickhouse:

https://gist.github.com/chancez/3da4b1df4f9942d2a260360a6a762912

chancez:
For testing: It shouldn't be terribly difficult to test using TLS, but it requires a bit of effort. I'd need some guidance on how you would expect certificates to be generated for testing.

In our project, we deploy clickhouse with TLS by using cert-manager in the cluster to issue certs. Alternatively, we could use openssl, though I'm not familiar with the python crypto APIs at all, so I'd prefer to just exec openssl to generate certificates for tests.

Once certs are generated, you just need to create a secret for them and then deploy using the example CHI resource I provided.

chancez:
Currently working on the tests. I have a minimal working example that I'm porting to tests atm.

chancez:
I "tried" to write some tests, but I'm having a hard time getting the test image to build via `./tests/image/build_docker.sh`, it just seems to fail in different places each time, sometimes pulling from quay, but most recently on:

```
Package docker-ce is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  docker-ce-cli:amd64
```

Also, it doesn't seem like the test image is multi-arch, so I might have issues regardless, as I'm on an m1 Mac.

I've pushed my attempt at some tests, but unfortunately it doesn't look like CI runs tests automatically, so I can't iterate that way either.

chancez:
Ah I just realized the reason the test image doesn't build is because it's using amd64 repos (and binaries) but I'm pulling the arm64 image. I guess I just have to update it to work for both.

Slach:
You can develop test without building image on arm64 platform

just setup `minikube`, `python3`, `python3-pip`, 'python3-venv`
and run
```bash
minikube start
python3 -m venv ~/operator-venv/
~/operator-venv/bin/pip3 install -U -r ./tests/image/requirements.txt
~/operator-venv/bin/python3 ./tests/regression.py --only="/regression/e2e.test_operator/yout_test*" --native
```
--native mode mean don't use docker-compose and use installed default minikube KUBECONFIG

chancez:
Here's the stack trace when I cancel it: it seems to be stuck on getting crds:

```
May 23,2022 11:02:50       ⟥  Given Clean namespace test, flags:MANDATORY
                11ms            [bash] 
                11ms            [bash] The default interactive shell is now zsh.
                11ms            [bash] To update your account to use zsh, please run `chsh -s /bin/zsh`.
                11ms            [bash] For more details, please visit https://support.apple.com/kb/HT208050.
                11ms            [bash] bash: kube_ps1: command not found
^C           10s 382ms       ⟥    Exception: Traceback (most recent call last):
                                    File "/Users/chancezibolski/projects/work/clickhouse-operator/./tests/regression.py", line 62, in <module>
                                      regression()
                                    File "/Users/chancezibolski/projects/work/clickhouse-operator/./tests/regression.py", line 55, in regression
                                      run_features()
                                    File "/Users/chancezibolski/projects/work/clickhouse-operator/./tests/regression.py", line 50, in run_features
                                      Feature(run=load(feature_name, "test"))
                                    File "/Users/chancezibolski/projects/work/clickhouse-operator/tests/e2e/test_operator.py", line 2210, in test
                                      util.clean_namespace(delete_chi=True)
                                    File "/Users/chancezibolski/projects/work/clickhouse-operator/tests/e2e/util.py", line 165, in clean_namespace
                                      kubectl.delete_all_chi(settings.test_namespace)
                                    File "/Users/chancezibolski/projects/work/clickhouse-operator/tests/e2e/kubectl.py", line 68, in delete_all_chi
                                      crds = launch("get crds -o=custom-columns=name:.metadata.name", ns=ns).splitlines()
                                    File "/Users/chancezibolski/projects/work/clickhouse-operator/tests/e2e/kubectl.py", line 40, in launch
                                      cmd = shell(cmd, timeout=timeout)
                                    File "/Users/chancezibolski/.asdf/installs/python/3.9.13/lib/python3.9/queue.py", line 180, in get
                                      self.not_empty.wait(remaining)
                                    File "/Users/chancezibolski/.asdf/installs/python/3.9.13/lib/python3.9/threading.py", line 316, in wait
                                      gotit = waiter.acquire(True, timeout)
                                  KeyboardInterrupt
```

Running `kubectl get crds -o=custom-columns=name:.metadata.name` on a fresh env results in just one line of output:
```
(⎈ |minikube:default) ~/p/w/clickhouse-operator ❯❯❯ kubectl get crds -o=custom-columns=name:.metadata.name                                secure_shard_communications ✭ ✱
name
```

chancez:
Ah, I thought the 

```
                13ms            [bash] bash: kube_ps1: command not found
```
wasn't a big deal, but running `bash` directly and trying to execute other commands shows that it's causing some issues. It seems like that error only comes up from bash, when I execute bash after sourcing my virtualenv, so the virtualenv is....persisting something and causing issues.

chancez:
Ok, I root caused it. It's expecting the prompt to match, so it's expecting the prompt to look correct, which is why things are busted. I was able to tweak some stuff to make it get passed this spot it's hanging, probably because of the ps1 issue I mentioned before.

chancez:
Alrighty, I got a local test setup working (built a custom docker image to run tests with) and got tests working. Please take a look.

Slach:
ok. look like your default shell is zsh
usually we run tests under bash in --native mode
@vzakaznikov JSFYI, as testflows author

chancez:
@Slach Yeah, it's the default on Mac, it seemed to be correctly opening bash, but I got it all sorted out. Running it in docker worked well.

Let me know if there's anything else this PR needs. I'm really going to need this so we can roll out Clickhouse with TLS in our environments.

chancez:
@Slach I want to be able to configure the cluster secret as a way to authenticate inter-node communication, eg: https://github.com/ClickHouse/ClickHouse/blob/b29e877f269e84ae452c446e70b406a695863470/tests/integration/test_distributed_inter_server_secret/configs/remote_servers_n1.xml#L4

Should that be a separate PR?

PR #1037, opened 2022-11-03
Adding test for volume reprovisioning.

antip00:
Adding test to check  that if volume is broken, operator re-creates volume and schema.


tests/e2e/test_operator.py
```diff
@@ -2907,6 +2908,80 @@ def test_035(self):
...
+        self.context.shell("""kubectl patch pv $(kubectl get pv | grep test/disk1-chi-test-036-volume-re-provisioning-simple-0-1-0 
+            | awk '{print $1}') -p '{"metadata":{"finalizers":null}}' """.replace('\r', '').replace('\n', ''))
+
+    with And("I check replica is deleted and recreated"):
+        replica_is_deleted = False
+        i = 0
+        with By("checking data from zookeeper table in cycle", description="I check replica 1 is disappeares and appeares in system.zookeeper table"):
+            while (i < num_retries) and not(replica_is_deleted and (r.output == "1")):
+                with When(f"try number {i}"):
+                    r = self.context.shell("""
```
alex-zaitsev, line 2957: Why not to use clickhouse.query() here?

tests/e2e/test_operator.py
```diff
@@ -2907,6 +2908,80 @@ def test_035(self):
...
+                        and name='chi-test-036-volume-re-provisioning-simple-0-1'" """.replace('\r', '').replace('\n', ''))
+                    if r.output == "0":
+                        replica_is_deleted=True
+                    i += 1
+                    time.sleep(5)
+
+        assert replica_is_deleted and (r.output == "1"), error()
+
+    with Then("I check PVC is recreated"):
+        r = kubectl.get_pvc_size("disk1-chi-test-036-volume-re-provisioning-simple-0-1-0")
```
alex-zaitsev, line 2971: use something like kubectl.wait_object() and wait_field() in order to check that PV is created back and the size is the same.

tests/e2e/test_operator.py
```diff
@@ -2907,6 +2908,80 @@ def test_035(self):
...
+        clickhouse.query(chi, f"INSERT INTO test_local_036 select * from numbers(10000)")
+
+    with When("I delete PV", description="delete PV on replica 1"):
+        pv_name = self.context.shell("kubectl get pv | grep test/disk1-chi-test-036-volume-re-provisioning-simple-0-1-0").output.split()[0]
+
+        self.context.shell(f"kubectl delete pv {pv_name} --force &")
+        self.context.shell("""kubectl patch pv $(kubectl get pv | grep test/disk1-chi-test-036-volume-re-provisioning-simple-0-1-0 
+            | awk '{print $1}') -p '{"metadata":{"finalizers":null}}' """.replace('\r', '').replace('\n', ''))
+
+    with And("I check replica is deleted and recreated"):
```
alex-zaitsev, line 2951: I think the check for ZooKeeper contents is not reliable in general. It is sufficient to test that PV is created back and data is properly replicated. It can not be done without handling replica properly in ZooKeeper anyway.

tests/e2e/test_operator.py
```diff
@@ -2907,6 +2908,80 @@ def test_035(self):
...
+            CREATE TABLE test_local_036 ON CLUSTER 'simple' (a UInt32)
+            Engine = ReplicatedMergeTree('/clickhouse/{installation}/tables/{shard}/{database}/{table}', '{replica}')
+            PARTITION BY tuple()
+            ORDER BY a
+            """.replace('\r', '').replace('\n', '')
+        clickhouse.query(chi, create_table)
+        clickhouse.query(chi, f"INSERT INTO test_local_036 select * from numbers(10000)")
+
+    with When("I delete PV", description="delete PV on replica 1"):
+        pv_name = self.context.shell("kubectl get pv | grep test/disk1-chi-test-036-volume-re-provisioning-simple-0-1-0").output.split()[0]
```
alex-zaitsev, line 2945: I would remove 0-0-0 PV instead of 0-1-0.

alex-zaitsev (changes requested):

antip00:
Changed and and created new PR.

PR #1044, opened 2022-11-09
Adding test for volume reprovisioning.

antip00:
Adding changed test for volume reprovisioning and updating requirements.


alex-zaitsev:
Since the test is supposed to fail now, add XFail to regression.py

antip00:
Xfail added, self.context.shell("kubectl get pv | grep test/disk1-chi-test-036-volume-re-provisioning-simple-0-0-0") changed to kubectl.get_pvc_size(f"disk1-chi-test-036-volume-re-provisioning-simple-0-0-0")

Issue #903, opened 2022-03-07
Automatic schema propagation does not propagate everything


macobo:
## Context

This chart mentions one feature:
> ClickHouse cluster scaling including **automatic schema propagation**

The behavior of this feature did not match my expectations.

## Current behavior

From @alex-zaitsev via Slack:

> Only Distributed tables and related local tables were created on a second shard [added after the fact]

This means tables which are not automatically propagated include:
- ReplicatedMergeTree tables _without_ Distributed tables (e.g. tables not using sharding)
- Kafka engine tables
- Materialized views
- Non-replicated tables

## Desired behavior

New shards should copy the full schema for everything affecting ingestion. In our usecase, this includes:

- ReplicatedMergeTree tables _without_ Distributed tables (e.g. tables not using sharding)
- Kafka engine tables
- Materialized views

Note wee'd be okay if other tables/whole schema got replicated.

As a minimum some capability to mark some tables as "to propagate" would be great.

macobo:
@alex-zaitsev reading the release notes of 0.19.0 I _think_ this can be closed? Haven't tested the behavior though.

alex-zaitsev:
It was fixed while ago

PR #1058, opened 2022-12-04
0.20.1 Adding test for automatic schema propapgating.

antip00:
test regarding issue https://github.com/Altinity/clickhouse-operator/issues/903.

test doesn't check every table engine except MaterializedPostgreSQL and RabbitMQ

test fails for now for live view and window view


tests/e2e/manifests/chi/test-038-1-schema-propagation.yaml
```diff
@@ -0,0 +1,76 @@
...
+    clusters:
+    - name: simple
+      schemaPolicy:
+        replica: All
+        shard: All
+      layout:
+        shardsCount: 1
+        replicasCount: 1
+    files:
+      graphite_rollup.xml: |
```
alex-zaitsev, line 22: Don't test Graphite, we do not support it

tests/e2e/manifests/chi/test-038-1-schema-propagation.yaml
```diff
@@ -0,0 +1,76 @@
...
+            - ReadWriteOnce
+          resources:
+            requests:
+              storage: 1Gi
+    podTemplates:
+      - name: multi-volume
+        spec:
+          containers:
+            - name: clickhouse-pod
+              image: clickhouse/clickhouse-server:22.8.6.71
```
alex-zaitsev, line 73: Use version from template.
```
  useTemplates:
    - name: clickhouse-version
```

alex-zaitsev (commented):
Good test, but it partially overlaps with test_013. 

I suggest to name it test_013_1, but add a few more changes but also test that all distribution policies are respected, including DistributedTablesOnly and None. Use one CHI for that to make it fast, just add extra steps that change schema settings for new shards.

Once the new test is merged I will remove old test_013.

alex-zaitsev (approved):

PR #1065, opened 2022-12-26
Adding test_038 Secure inter-cluster communication.

antip00:
https://altinity.atlassian.net/browse/CHO-402


alex-zaitsev (changes requested):
1. Test should check that distributed table is created AND that user can select from it. 'secure' flag turns on intercluster HTTPS, that work during selects
2. 'default' user should by set a custom password first, and run a test w/o secure -- the test is expected to fail
3. After that check with 3 different secret configuration for 'secure' mode -- it should pass.
4. It also makes sense to inspect remote_servers.xml in a configmap, and check that secret and secure flags are in there
5. hostTemplate should not be used. Please rebase to 0.20.2 -- 'secure' flag should be in 'cluster' section. It did not work correctly before 0.20.2

PR #1075, opened 2023-01-10
0.20.2 Adding tests for secure inter-cluster communications.

antip00:
Adding tests for secure inter-cluster communications.


alex-zaitsev:
Moved to test-039-4-communications-with-secret.yaml

PR #1085, opened 2023-01-30
0.20.2 reformatting.

antip00:
Moving getting shell to a separate step. Reformatting with black reformater.


Slach (changes requested):
could you reformat with right margin 120 instead of right marging 80 which have no sense ;)

Slach (approved):

Issue #1109, opened 2023-02-14
Configurable concurrency level in reconciliation of shards


zcross:
Has anybody previously looked into modifying the worker implementation to support concurrent reconciliation of more than one shard at once? I have done a little searching in the project's issues and have only come across #376 which seems to be based on the same concern (time of operations on large clusters), though that's limited to creation and I'm interested in reconciliation in general.

Given sequential reconciliation over a cluster's shards, in which each statefulset pod takes O(single digit minutes) to complete, large clusters can take hours to reconcile. When the operation in question is some action needed to restore stability, that translates to long time to resolution (and in turn, potentially long periods of sustained negative impact).

My few concrete ideas for now:
- Orchestrate the sequential work in [this for loop over shards](https://github.com/Altinity/clickhouse-operator/blob/master/pkg/apis/clickhouse.altinity.com/v1/type_chi.go#L419) with a configurably-sized worker pool (this is extremely reductive and probably overlooking lots of other internal state)
- Add simple synchronization to the underlying map behind the `Registry` type (which I *think* is used within the scope of a single worker) – as well as all other internal mutable state used during shard reconciliation
- To avoid opting all current users of the operator into this behavior, make it configurable and use a default matching the status quo (n=1 concurrency, i.e., no additional concurrency)
- Use the fact that the shard layout is static to support either absolute number specification of concurrency, or a percentage specification. Both should have semantics similar to that of k8s' `MaxUnavailable` option of rolling updates (which I realize is not applicable to the stateful sets modeling each replica of each shard), i.e., phrasing how many shards will be affected in parallel (even though we should try to maintain the availability of any shards that use replication by reconciling a replica/host at a time).
- While this may not match every usage pattern of clickhouse, it should be a useful opt-in behavior for some use cases. 
- I'm not proposing adding intra-shard concurrency (e.g., I think we want to maintain the availability of at least one replica within each shard at all times, in configurations using replication).

[...]

PR #1115, opened 2023-02-24
Refactor Registry internals to make concurrent access safe

zcross:
Background: #1109

Related PR(s):
* #1119
* #1124

This adds simple synchronization (`sync.RWMutex`) to accesses of the mutable state within `Registry`. 
It also restructures the data to use a map per entity type, rather than a slice, replacing some linear searches of the data.
As recommended by @sunsingerus , I'm starting here with the Registry and planning on moving onto CHI status (and probably the announcer, at least with a slight change to its builder pattern for chaining log line inputs).

For me, the most pressing open question is whether my read of the code _using_ Registry is semi-accurate: it seems that it is used for internal book keeping, and that individual `ObjectMeta` instances are read in the `Walk...` functions without direct mutation. As I write that, I'm considering whether that was part of the intent of the original code passing these structs by value rather than by pointer. 

Meta: I ran into some issues executing the e2e tests locally (have not fixed yet) and some other simpler (fixed) problems developing this branch on an M1 Mac. I don't want to pull the fixes I made locally into the discussion here, but I won't be shocked if the e2e tests reveal some issue I'm not yet aware of due to this dev environment issue.

### DIFF
--- a/pkg/model/registry.go
+++ b/pkg/model/registry.go
@@ -16,6 +16,7 @@ package model
 
 import (
 	"fmt"
+	"sync"
 
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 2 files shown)


pkg/model/registry.go
```diff
@@ -293,48 +335,99 @@ func (r *Registry) isEqual(a, b v1.ObjectMeta) bool {
...
+	// and then begin operating on that (it uses a finer grained lock).
+	setForType := r.getOrCreateObjSet(entityType)
+
+	// Having acquired the type-specific ObjectMeta set, return the result of removal success.
+	return setForType.remove(&meta)
+}
+
+// getOrCreateObjSet resolves the singleton objectMetaSet for this registry, of the given entityType.
+// This uses a coarse grained lock on the entire registry.
+func (r *Registry) getOrCreateObjSet(entityType EntityType) *objectMetaSet {
```
sunsingerus, line 348: We have an internal agreement that functions with behavior like 'get or create if not exists and then get' are called either `Ensure()` or starting with `Ensure`, like `Ensure<Something>()`
Ex.:
```
func (chi *ClickHouseInstallation) EnsureStatus() *ChiStatus {
```
or
```
func (o *StatefulSetPollOptions) Ensure() *StatefulSetPollOptions {
```
Most likely this is the case

sunsingerus (changes requested):

zcross:
> > ... Registry ... is used for internal book keeping, and that individual ObjectMeta instances are read ... without direct mutation
> 
> Yes, this is correct.
> 
> Main idea of the `Registry` is to keep registry of reconciled objects for two main reasons (at least for now)
> 
>     1. To clear (delete) all objects that belong to the CHI but not listed in registry as reconciled - thus removing all remnants of previous iterations of the `CHI` w/o knowing what they are. Clean all what is not listed in the registry as belonging to current iteration of the `CHI`
> 
>     2. Statistics and counting
> 
> 
> Entities inside `Registry` are not mutated, but are read only, at least for now

Perfect, I think that's compatible with the approach taken in this PR in that case.

pkg/model/registry.go
zcross: (Addressed)

zcross:
Meta: I understand that you want commits squashed. I tend to incrementally commit to address PR feedback, so that the feedback-inspired changes are visible at the commit level while the PR is under review, and then squashing to one commit before merging. Does that work, or do you prefer squashing aggressively during the review process? Thanks!

sunsingerus:
> do you prefer squashing aggressively during the review process?

No need to squash immediately, this can be done once at the very end

sunsingerus:
> ... project seems to rely on end to end behavioral tests. ...

The intention is to cover the whole functionality (or at least as much as possible) with e2e tests. Tests are typically added along with new functionality. However, in case any strange / corner case behavior is found later, tests are modified to cover it. Generally speaking, successful run of the test suite gives good hope that the operator can do most of the typical use-case scenarios.

> ... I was wondering about dropping in some very simple go tests ...

Yes, sure. Tests covering any functionality (new and already existing, and especially complex scenarios) are always welcome.

pkg/model/registry_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+var testPvcD = v1.ObjectMeta{
+	Name:      "persistentvolumeclaim-D",
+	Namespace: testNamespace,
+}
+
+// NB: These tests mostly exist to exercise synchronization and detect regressions related to them via the
+// Golang race detector. See: https://go.dev/blog/race-detector
+// In short, add -race to the go test flags when running this.
+func Test_Registry_BasicOperations_ConcurrencyTest(t *testing.T) {
```
zcross, line 54: @sunsingerus here's the minimal (not assertion-heavy) go test I mentioned, its purpose being to exercise the synchronization primitives and detect any regressions related to concurrency

pkg/model/registry.go
```diff
@@ -293,48 +335,99 @@ func (r *Registry) isEqual(a, b v1.ObjectMeta) bool {
...
+	r.RUnlock()
+	if ok {
+		return existing
 	}
 
-	return false
+	// create case (requires write lock)
+	newSet := newObjectMetaSet(entityType)
+	r.Lock()
+	r.r[entityType] = newSet
```
zcross, line 364: this should actually check for existence again, will fix

zcross:
Squashing this as I await feedback and repointing the other open PR to this as its base so I can branch a third stacked PR for the actual reconciliation logic changes.

PR #1119, opened 2023-03-02
Refactor ChiStatus internals to make concurrent access safe

zcross:
Background: https://github.com/Altinity/clickhouse-operator/issues/1109
Related PR(s):
* #1115
* #1124

This PR adds `sync.RWMutex` based synchronization over the mutable state of `ChiStatus`. It should be relatively straightforward. (Also, around the initialization of `ClickHouseInstallation`'s own status field via `EnsureStatus`).

The most eyebrow raising bits are probably in the sequenced lock acquisition in `MergeActions` and `CopyFrom` creating some risk of deadlocking (to be discussed, I'm sure). The only other potentially "hairy" concern is some of the generated code I've found for `DeepCopy` helpers on all types. 

Meta: I began this by trying to define a mediator through which all updates would be made. However, the presence of logic inside some of the other raw types inside of the `v1` package (such as the `ClickHouseInstallation` AKA chi) made it apparent to me that it's not that much cleaner to separate the two. My workaround to this was to at least make sure that all field level read and write accesses to the `ChiStatus` struct were strictly within its single source file (`type_status`) – with the exception of the deepcopy generated files (which don't seem to be used directly AFAIK, and hopefully aren't used at runtime from potentially competing goroutines).

Note: I'll squash all commits from feedback pre-merge.

* [X] The PR is made into dedicated `next-release` branch, **not into** `master` branch<sup>1</sup>. [More info](https://github.com/Altinity/clickhouse-operator/blob/master/CONTRIBUTING.md#how-to-make-pr)
* [X] The PR is signed. [More info](https://github.com/Altinity/clickhouse-operator/blob/master/CONTRIBUTING.md#sign-your-work)

### DIFF
--- a/pkg/apis/clickhouse.altinity.com/v1/type_chi.go
+++ b/pkg/apis/clickhouse.altinity.com/v1/type_chi.go
@@ -25,30 +25,28 @@ import (
 	"gopkg.in/yaml.v3"
 
 	"github.com/altinity/clickhouse-operator/pkg/util"
-	"github.com/altinity/clickhouse-operator/pkg/version"
 )
 
 // FillStatus fills .Status
 func (chi *ClickHouseInstallation) FillStatus(endpoint string, pods, fqdns []string, ip string) {
-	chi.EnsureStatus().CHOpVersion = version.Version
-	chi.EnsureStatus().CHOpCommit = version.GitSHA
-	chi.EnsureStatus().CHOpDate = version.BuiltAt
-	chi.EnsureStatus().CHOpIP = ip
-	chi.EnsureStatus().ClustersCount = chi.ClustersCount()
-	chi.EnsureStatus().ShardsCount = chi.ShardsCount()
-	chi.EnsureStatus().HostsCount = chi.HostsCount()
-	chi.EnsureStatus().TaskID = chi.Spec.GetTaskID()
-	chi.EnsureStatus().HostsUpdatedCount = 0
-	chi.EnsureStatus().HostsAddedCount = 0
-	chi.EnsureStatus().HostsCompletedCount = 0
-	chi.EnsureStatus().HostsDeleteCount = 0
-	chi.EnsureStatus().HostsDeletedCount = 0
-	chi.EnsureStatus().Pods = pods
-	chi.EnsureStatus().FQDNs = fqdns
-	chi.EnsureStatus().Endpoint = endpoint
-	chi.EnsureStatus().NormalizedCHI = chi.Copy(CopyCHIOptions{
-		SkipStatus:        true,
-		SkipManagedFields: true,
+	chi.EnsureStatus().Fill(&FillStatusParams{
+		CHOpIP:              ip,
+		ClustersCount:       chi.ClustersCount(),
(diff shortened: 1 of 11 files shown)


zcross:
FYI @sunsingerus – apologies for closing the first attempt at this `ChiStatus` refactor, but I force pushed it without realizing that would prevent re-opening.

zcross:
Hmm, I just began a branch to refactor `Announcer` (and its variations like `chi.Announcer` and the others). However, It seems to me like this code is all already goroutine safe as all of the "fluent style" (chained) function calls operate on and return struct values rather than pointers, i.e., beginning with a copy by value before mutating any state:

```
b := a
b.SomeField = newLogCallSpecificValue
return b
```

Seem like a correct assessment? If so, there won't really be an "announcer PR" – I'll be turning to a final (third) PR to actually update reconciliation logic (well, returning to my original draft, basing it off of this and the registry branches, and then cleaning it up for PR-readiness)

pkg/apis/clickhouse.altinity.com/v1/type_chi.go
```diff
@@ -25,28 +25,26 @@ import (
 	"gopkg.in/yaml.v3"
 
 	"github.com/altinity/clickhouse-operator/pkg/util"
-	"github.com/altinity/clickhouse-operator/pkg/version"
 )
 
 // FillStatus fills .Status
 func (chi *ClickHouseInstallation) FillStatus(endpoint string, pods, fqdns []string, ip string) {
-	chi.EnsureStatus().CHOpVersion = version.Version
```
zcross, line 33: version is always statically available so these 3 field setter calls were pushed within the new `Fill` function (and therefore absent from the passed params)

sunsingerus:
Code-generator produces errors on generics introduced in this PR. This issue blocks merge atm and needed to be looked into.

sunsingerus:
@zcross merge is blocked atm until further understanding on code-generator + generics situation.

zcross:
Ok, ACK that I saw this and will take a shot at cleaning it up next week. Re: generics, I'll just replace them with a few type specific functions as needed for simplicity. BTW: how can I reproduce the problem you're seeing for my own iteration? (E.g., what make/build command are you running) Thanks!

sunsingerus:
`clientset`, `informer` and other code is generated with `./dev/run_code_generator.sh ` command
```
clickhouse-operator$ ./dev/run_code_generator.sh 
...
Generating clientset for clickhouse.altinity.com:v1 at github.com/altinity/clickhouse-operator/pkg/client/clientset
W0414 17:07:52.225131 2075270 parse.go:818] Making unsupported type entry "T" for: &types.TypeParam{check:(*types.Checker)(nil), id:0x2, obj:(*types.TypeName)(0xc006262780), index:0, bound:(*types.Interface)(0xc0000bcaa0)}
...
```

zcross:
Sorry for not getting to this until today, but I just:
* Resolved merge conflicts (fairly straightforward)
* Replaced generics (which I didn't notice were the first generic usage in this repo) with type specific equivalents – only string/int/installation pointer types were needed so no big deal

PR #1171, opened 2023-06-12
Tests paralleling and test 041.

antip00:
Making tests running in parallel and adding test 041 for secure zookeeper.


deploy/zookeeper/quick-start-persistent-volume/zookeeper-1-node-secure-for-test.yaml
```diff
@@ -0,0 +1,307 @@
+# Setup Service to provide access to Zookeeper for clients
```
Slach, line 1: Please consider using `zookeeper-1-node-1GB-for-tests-only-scaleout-pvc.yaml` as a template

Slach (changes requested):

tests/e2e/test_operator.py
```diff
@@ -3442,21 +3588,85 @@ def test_040(self):
...
+    with Then("I check clickhouse can successfully connect to zookeeper"):
+        clickhouse.query(chi, "SELECT * FROM system.zookeeper WHERE path = '/'")
+
+    with And("I check data is distributed"):
+        r = clickhouse.query(chi, "SELECT count(*) FROM secure_dist")
+        assert r == "10"
+
+    with And("I check connection is secured"):
+        with By("checking chop-generated-zookeeper.xml is properly configured"):
+            r = kubectl.launch(f'exec chi-{chi}-default-0-0-0 -- bash -c "cat '
```
Slach, line 3643: instead of tail

use `cat /etc/clickhouse-server/conf.d/chop-generated-zookeeper.xml | grep -c "<secure>1</secure>"`

Slach (changes requested):

Slach (approved):

PR #1178, opened 2023-06-22
Fixes for parallel tests.

antip00:
2 fixes for tests:

1. Test namespace is created only if it is needed.
2. If test is failed namespace is not deleted.

### DIFF
--- a/tests/e2e/test_operator.py
+++ b/tests/e2e/test_operator.py
@@ -35,6 +35,12 @@ def test_001(self):
             "pdb": ["single"],
         },
     )
+    with Then("I delete namespace"):
+        shell = get_shell()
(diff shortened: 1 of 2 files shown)


tests/e2e/test_operator.py
```diff
@@ -35,6 +35,12 @@ def test_001(self):
             "pdb": ["single"],
         },
     )
+    with Then("I delete namespace"):
```
Slach, line 38: if we have create_test_namespace()

why we make lot of copy\paste code instead of create
delete_test_namespace() ?

Slach (changes requested):

Slach (approved):

Issue #1187, opened 2023-07-06
Nil pointer dereference in metrics exporter


zcross:
I recently observed this on `0.21.2` while bootstrapping a brand new installation. FWIW, it was "self healing" after a few minutes: ongoing reconciliation operations seemed to bring the installation(s) into a state that caused the status to no longer be nil.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x198 pc=0x1370d8e] goroutine 1 [running]:
github.com/altinity/clickhouse-operator/pkg/apis/metrics.(*Exporter).DiscoveryWatchedCHIs(0x16c6661?, {0x18f6338, 0xc00002e6e0}, 0xc0000317f0) /clickhouse-operator/pkg/apis/metrics/exporter.go:295 +0x3ee
github.com/altinity/clickhouse-operator/cmd/metrics_exporter/app.Run() /clickhouse-operator/cmd/metrics_exporter/app/metrics_exporter.go:108 +0x36b
main.main() /clickhouse-operator/cmd/metrics_exporter/main.go:23 +0x17
```
https://github.com/Altinity/clickhouse-operator/blob/32ef0fadc3d07884845a1ee3d91be6b6e112ac62/pkg/apis/metrics/exporter.go#L295

My guess is that we can fix this by changing the access of `chi.Status.NormalizedCHICompleted` to `chi.EnsureStatus().GetNormalizedCHICompleted()`, or something like that. I can't do it right this moment, but I'll circle back to this issue when I get some bandwidth.

To generalize: maybe we want to use Golang visibility to make raw/unsynchronized fields on `Status` package-internal to more strongly encourage usage of safe public getter/setter APIs. I considered this in #1119 but IIRC, this resulted in problems when trying to deserialize/serialize the state of `Status` (because it relied on field visibility)... I think?

zcross:
Let's close this if the build with #1188 passes? I'm not sure how deterministic it is to repro, but one thing worth trying is just bringing up a fresh installation and seeing if that reproduces on `0.21.2` – then trying again with `0.21.3`

alex-zaitsev:
@zcross , this is fixed in 0.21.3

zcross:
Thank you!

zcross:
> @zcross , this is fixed in 0.21.3

Would it be possible to cut a release tag for `0.21.3` to distribute this fix?

PR #1188, opened 2023-07-07
Fix nil pointer deref in metrics exporter (#1187) 

zcross:
This is a very targeted fix for the NPE shown in #1187 since it's a known problem. The general pattern of guarding status lookups may be worth thinking about. Maybe we just need an intermediate type that decouples the application level usage of this state from its serialization/deserialization, at which point we can use basic language features like field visibility to reduce the probability of future unsafe accesses. 

* [ ] All commits in the PR are squashed. [More info](https://github.com/Altinity/clickhouse-operator/blob/master/CONTRIBUTING.md#how-to-make-pr) 
* [ ] The PR is made into dedicated `next-release` branch, **not into** `master` branch<sup>1</sup>. [More info](https://github.com/Altinity/clickhouse-operator/blob/master/CONTRIBUTING.md#how-to-make-pr)
* [ ] The PR is signed. [More info](https://github.com/Altinity/clickhouse-operator/blob/master/CONTRIBUTING.md#sign-your-work)

### DIFF
--- a/pkg/apis/metrics/exporter.go
+++ b/pkg/apis/metrics/exporter.go
@@ -292,7 +292,7 @@ func (e *Exporter) DiscoveryWatchedCHIs(kubeClient kube.Interface, chopClient *c
 			log.V(1).Infof("Skip stopped CHI %s/%s\n", chi.Namespace, chi.Name)
 		} else {
 			log.V(1).Infof("Add explicitly found CHI %s/%s", chi.Namespace, chi.Name)
-			if chi.Status.NormalizedCHICompleted == nil {
+			if !chi.EnsureStatus().GetNormalizedCHICompleted() {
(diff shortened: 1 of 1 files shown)


zcross:
Meta: it would seem to be going against the trend to add some highly specific unit test for this, so I'm going to let integration tests "cover it" ... although they didn't catch its first introduction (possibly due to the self-healing / temporal element of the "status" state).

sunsingerus (approved):

zcross:
Whoops: 

```
133.0 pkg/apis/metrics/exporter.go:295:8: invalid operation: operator ! not defined on chi.EnsureStatus().GetNormalizedCHICompleted() (value of type *"github.com/***/clickhouse-operator/pkg/apis/clickhouse.***.com/v1".ClickHouseInstallation)
```

I'll fire this up on my actual development machine and fix that now

PR #1189, opened 2023-07-07
Fix syntax issue introduced in #1188 NPE bugfix

zcross:
While manually writing the fix in #1188 using a text editor, I misread the return type of the `GetNormalizedCHICompleted` function as returning a bool rather than a pointer to a CHI. That broke the build of `0.21.3` once #1188 was merged.

Now, this looks more similar to the original state of the code (just using getters instead of direct field access):
https://github.com/Altinity/clickhouse-operator/pull/1188/files#diff-6dd1bf383889aa0b3dde98cf3579859d84b90dd638c4829390596f9237e62f65L295

cc @sunsingerus 

### DIFF
--- a/pkg/apis/metrics/exporter.go
+++ b/pkg/apis/metrics/exporter.go
@@ -295,7 +295,7 @@ func (e *Exporter) DiscoveryWatchedCHIs(kubeClient kube.Interface, chopClient *c
 			continue
 		}
 
-		if !chi.GetStatus().HasNormalizedCHICompleted() {
+		if !chi.EnsureStatus().HasNormalizedCHICompleted() {
 			log.V(1).Infof("CHI %s/%s is not completed yet, skip it\n", chi.Namespace, chi.Name)
 			continue
 		}


zcross:
Force pushed to fix conflict (`DiscoveryWatchedCHIs` had some changes on `0.21.3` after my unfortunately syntactically-broken first PR)

pkg/apis/metrics/exporter.go
```diff
@@ -295,7 +295,7 @@ func (e *Exporter) DiscoveryWatchedCHIs(kubeClient kube.Interface, chopClient *c
 			continue
 		}
 
-		if !chi.GetStatus().HasNormalizedCHICompleted() {
```
zcross, line 298: `GetStatus` can return nil, so we could either:
* `if chi.GetStatus() == nil || !chi.GetStatus().HasNormalizedCHICompleted()`
* `if chi.EnsureStatus().HasNormalizedCHICompleted()`

I picked the second option for brevity (and because I'm thinking there's no harm to just ensuring that the status struct is non-nil on access like this, i.e., because it is synchronized and can't "clobber" any other concurrent mutations to this state).

alex-zaitsev:
The problem is already fixed, so this PR is not needed anymore

PR #1192, opened 2023-07-13
Fix shard weight. Now we can set it to zero

maxistua:
I can't set the shard weight to zero
We need it for some reasons...

Hopefully this MR will fix that


UnamedRus:
It's problem with other settings as well AFAIK.
If value is 0, it will not be materialized in xml config by operator

alex-zaitsev (changes requested):
Please provide a usage example. Ideally a test, but even example would be helpful.

sunsingerus:
I'll take a look on this PR

PR #1234, opened 2023-09-04
Improve clickhouse-keeper manifests

Slach:
Depends on https://github.com/ClickHouse/ClickHouse/pull/53481

This will help a lot for maintainers to adopt your Pull Request. 

* [ ] All commits in the PR are squashed. [More info](https://github.com/Altinity/clickhouse-operator/blob/master/CONTRIBUTING.md#how-to-make-pr) 

--

<sup>1</sup> If you feel your PR does not affect any Go-code or any testable functionality (for example, PR contains docs only or supplementary materials), PR can be made into `master` branch, but it has to be confirmed by project's maintainer.


Slach:
Wait when  resolve https://github.com/ClickHouse/ClickHouse/pull/53481 and https://github.com/ClickHouse/ClickHouse/issues/54129

bputt-e:
does this fix the issue where `force-recovery` was always set to the first pod if I'm running a 3 clickkeeper cluster? It seems odd that `force-recovery` was always set to the first pod as if that pod were restarted or offline for a period of time, it effectively tells the rest of the clickkeeper instances that they must use the first pods state

Slach:
using `--force-recovery` in old manifest required for prefer XML configs have more priority then internal NuRaft state which stored in /var/lib/clickhouse-keeper, without this option even when you change XML config <raft_configuration> it will have no effect

in the new approach we trying use `reconfig` command, which stay not fully operable

PR #1300, opened 2024-01-05
ADD the user migration function when clickhouse expanding capacity.

ccsxs:
The function of synchronizing user information from the original host to the new host when clickhouse is added. First, it is judged whether the current host is working and whether it is added. Start user synchronization operation for new host. Select a working host from the original cluster, which is different from the current host, and name it select host. Compare the user information in the current host and the select host, and if they are different, perform user synchronization operation, and copy the /var/lib/clickhouse/access directory from the select host to the current host. Then execute the reload user information command “system reload users” in clickhouse-client to synchronize user information.

### DIFF
--- a/pkg/controller/chi/exec.go
+++ b/pkg/controller/chi/exec.go
@@ -1,0 +1,71 @@
+package chi
+
+import (
+	"fmt"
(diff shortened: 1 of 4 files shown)


ccsxs:
resolve the conflict

PR #1311, opened 2024-01-25
neat: remove dangling `data` property

ianaya89:
- Remove unused `data` property from deploy manifest
- Remove dangling spaces

### DIFF
--- a/deploy/operator/clickhouse-operator-install-bundle-v1beta1.yaml
+++ b/deploy/operator/clickhouse-operator-install-bundle-v1beta1.yaml
@@ -3344,7 +3344,6 @@ metadata:
   labels:
     clickhouse.altinity.com/chop: 0.22.2
     app: clickhouse-operator
-data:
 ---
 # Template Parameters:
 #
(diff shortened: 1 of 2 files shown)


Slach:
thanks for contribution, unfortunatelly this PR do nothing

you changed generated file
look to dev folder and try to figure out how exactly it works

Why you struggle with emtpy "data" section?

deploy/operator/clickhouse-operator-install-bundle-v1beta1.yaml
```diff
@@ -3344,7 +3344,6 @@ metadata:
   labels:
     clickhouse.altinity.com/chop: 0.22.2
     app: clickhouse-operator
-data:
```
Slach, line 3347: why? which effect?

Slach (changes requested):

deploy/operator/clickhouse-operator-install-bundle-v1beta1.yaml
ianaya89: @Slach thanks for clarifying. Is not an actual issue, but the property looks to be empty (nothing declared inside).
Does not cause problems other than DX (get false warning for linters or that kind of stuff).

Slach:
let's close PR

Issue #1138, opened 2023-03-30
Configure watching other namespaces via helm chart


chancez:
Currently I'm trying to use the new helm chart and I'm installing clickhouse-operator into it's own `clickhouse-operator` namespace, but `watchNamespaces: []` in the configmap results in it only watching it's own namespace. I don't want to have to completely overwrite the config file, just adjust the namespaces to be watched. Is this currently possible?

Slach:
only one way,

copy \ paste
```yanl
files:
  files:
    config.yaml: |-
...
```
from values.yaml

and change to
```yaml
watch:
  namespaces: ["other-namespace"}
```

or 
```yaml
watch:
  namespaces: [""}
```
if you want to watch all namespaces

chancez:
The file contains a lot of other stuff that can change between versions, so that's not going to be maintainable for downstream users.

itssimon:
Wondering the same here. Copy & pasting the entire config file just to adjust one value won't work for us either. There must be a better way?

echozio:
The value of configs.files.* are templated, so adding some interpolation for commonly configured parameters would make this easier to deal with:
```yaml
watch:
  namespaces: ["some-namespace", ".*"]

configs:
  files:
    config.yaml: |
      ...
      watch:
        namespaces: {{ default (list) .Values.watch.namespaces | toJson }}
      ...
```
An alternative would be to keep the contents as yaml rather than a string, so that every user who needs to change one small thing doesn't need to maintain and keep up-to date a config they've barely changed.
This would do that and be backwards-compatible:
```diff
diff --git a/deploy/helm/templates/generated/ConfigMap-etc-clickhouse-operator-files.yaml b/deploy/helm/templates/generated/ConfigMap-etc-clickhouse-operator-files.yaml
index b8f3491d..d7228798 100644
--- a/deploy/helm/templates/generated/ConfigMap-etc-clickhouse-operator-files.yaml
+++ b/deploy/helm/templates/generated/ConfigMap-etc-clickhouse-operator-files.yaml
@@ -10,4 +10,10 @@ metadata:
   name: {{ printf "%s-files" (include "altinity-clickhouse-operator.fullname" .) }}
   namespace: {{ .Release.Namespace }}
   labels: {{ include "altinity-clickhouse-operator.labels" . | nindent 4 }}
-data: {{ tpl (toYaml .Values.configs.files) . | nindent 2 }}
+data:
[...]

PR #1317, opened 2024-01-28
update helm chart generator to treat config.yaml as yaml in values

echozio:
This addresses some of the concerns raised in #1138 and leaves the contents of `config.yaml` as YAML in `values.yaml`, enabling users to change a single value without the need to copy the rest of its contents over to their own values.

Users already using a string for their `config.yaml` value should not be affected, as the template partial responsible for rendering the data treats any top-level string value as before. This behavior was also necessary to retain in order to render the ConfigMaps with non-YAML data.

---


Slach:
thanks for contribution
hope not so much users used `config.yml` provided as string

PR #1333, opened 2024-02-05
Updated grants example

lesandie:
Updated documentation grants example with GRANT role.


Slach (approved):

Slach:
thanks for contribution

PR #1360, opened 2024-03-03
Changed Merge metric from ProfileEvent to CurrentMetric

lesandie:
To have a more detailed count of current merges in a particular point in time, changed the ProfileEvent rate function to a CurrentMetric. New Altinity Monitoring dashboard (@filimonov) uses CurrentMetric to measure merges instead of using a ProfileEvent with a rate.


Slach:
these changes have not sense and will less preciese
we need Merges/sec, instead of "merges on the fly"
look to
https://fiddle.clickhouse.com/998761fd-8cf4-44b1-9287-8dc20512d182
12 much better than 0

PR #1365, opened 2024-03-07
docs: Updated ClickhouseKeeper example with securityContext

vitaliyf:
Without securityContext I was getting "permission denied" writing to /var/lib/c-k/coordination/ directories.


Slach:
Thanks for contributing

PR #1409, opened 2024-04-30
clustergen: add all-clusters cluster

### LINKED ISSUE #1291: all-shards cluster
We have a use case where we would like to have multiple CH clusters, but allow querying all shards from each cluster from a single place, similar to `all-sharded`. The key difference being that we respect the replication of shards, instead of treating each replica as its own shard.

For example:

```yaml
clusters:
  prod_one:
    replicas: 2
    shards: 2
  prod_two:
    replicas: 2
    shards: 2
```

and have a resulting settings.conf cluster like

```xml
<clusters>
  <all-shards>
    <shard>
      <internal_replication>True</internal_replication>
      <replica>
        prod_one-0-0 // shorthand for brevity
      </replica>
      <replica>
        prod_one-0-1
      </replica>
    </shard>
    <shard>
      <internal_replication>True</internal_replication>
      <replica>
        prod_one-1-0
      </replica>
      <replica>
        prod_one-1-1
      </replica>
    </shard>
    <shard>
      <internal_replication>True</internal_replication>
      <replica>
        prod_two-0-0
      </replica>
      <replica>
        prod_two-0-1
      </replica>
    </shard>
    <shard>
      <internal_replication>True</internal_replication>
      <replica>
        prod_two-1-0
      </replica>
      <replica>
        prod_two-1-1
      </replica>
    </shard>
  </all-shards>
</clusters>
```

We could then create a distributed table on the `all-shards` cluster, which would fan out the query to each shard in all clusters.

I have a draft of the code required to do so already. If a PR for this is welcome, I will happily make one

tanner-bruce:
Adds an `all-shards` cluster, which contains the shard definitions for each cluster specified in the ClickhouseInstallation. This is used by our `query` nodes to allow scattering reads across all clusters in the CHI while respecting the shard/replica.

fixes #1291

### DIFF
--- a/pkg/model/chi/ch_config_generator.go
+++ b/pkg/model/chi/ch_config_generator.go
@@ -33,6 +33,7 @@ const (
 	// 2. Cluster with all shards (1 replica). Used to gather/scatter data over all replicas.
 	OneShardAllReplicasClusterName = "all-replicated"
 	AllShardsOneReplicaClusterName = "all-sharded"
+	AllClustersClusterName         = "all-clusters"
 )
 
 // ClickHouseConfigGenerator generates ClickHouse configuration files content for specified CHI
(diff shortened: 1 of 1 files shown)


Slach (changes requested):
You are ignored our recommendations about non-mental name `all-shards`
https://github.com/Altinity/clickhouse-operator/issues/1291#issuecomment-1855192960
could you please use another name like `all-clusters`?

Slach:
Or even `all-clusters-shards`

tanner-bruce:
> You are ignored our recommendations about non-mental name `all-shards` [#1291 (comment)](https://github.com/Altinity/clickhouse-operator/issues/1291#issuecomment-1855192960) could you please use another name like `all-clusters`?

Let me check what happened, I pushed the wrong code. We have it as `chi-<cluster>`

tanner-bruce:
> You are ignored our recommendations about non-mental name `all-shards` [#1291 (comment)](https://github.com/Altinity/clickhouse-operator/issues/1291#issuecomment-1855192960) could you please use another name like `all-clusters`?

Fixed now. I've pushed the proper code, which changed the name to what you suggested in #1291 - `chi-<cluster name>`.  Please let me know if you'd prefer something else. Or maybe we could make it configurable?

Slach (approved):

pkg/model/chi/ch_config_generator.go
```diff
@@ -31,8 +31,9 @@ const (
 	// Special auto-generated clusters. Each of these clusters lay over all replicas in CHI
 	// 1. Cluster with one shard and all replicas. Used to duplicate data over all replicas.
 	// 2. Cluster with all shards (1 replica). Used to gather/scatter data over all replicas.
-	OneShardAllReplicasClusterName = "all-replicated"
-	AllShardsOneReplicaClusterName = "all-sharded"
+	OneShardAllReplicasClusterName  = "all-replicated"
+	AllShardsOneReplicaClusterName  = "all-sharded"
+	AllShardsAllClustersClusterName = "chi-%s"
```
Slach, line 36: After internal discussion let's make it as
```
AllClustersAllShardsAndReplicasClusterName = "all-clusters"
```

Slach (changes requested):

pkg/model/chi/ch_config_generator.go
tanner-bruce: Updated.

PR #1425, opened 2024-05-23
0.24.0 keeper upgrade

antip00:
Adding test to check clickhouse-keeper upgrade.
This test fails due to the fact that local tables switch to a read-only mode.

Adding check to test_048, that pvc is supported for ClickHouseKeeperInstallation.


hodgesrm:
We're seeing this issue in the field. Is there a workaround for tables going R/O when Keeper is upgraded?

sunsingerus (approved):

PR #1457, opened 2024-07-15
chore: use ConfigMap instead of Secret for grafana dashboards

cwrau:
secrets should only be used for sensitive data, which dashboards are not


Slach:
yep, according to https://github.com/grafana/helm-charts/blob/grafana-8.3.4/charts/grafana/values.yaml#L778
it should be `configMap` instead of secrets

Issue #1433, opened 2024-06-07
Able to pause ClickHouseInstallation object reconcilation for maintenance mode


dmvolod:
**Motivation**
For some ClickHouse technical maintenance it would be nice to have an option to pause reconcilation and avoiding updates of the related Kubernetes objects.

**Solution**
For this implementation, `clickhouse.altinity.com/paused` annotation can be introduced and even if it present, any CRUD operation on the ClickHouseInstallation object and related objects should be skipped while reconcilation.

alex-zaitsev:
This is a good idea. Added to 0.25 release planning

alex-zaitsev:
@dmvolod , as a workaround, you may scale down operator deployment to 0 replicas.

dmvolod:
> @dmvolod , as a workaround, you may scale down operator deployment to 0 replicas.

Thanks for suggestion, @alex-zaitsev , I know about it, but in multi-tenant and strong secured environments it's a bit difficult to implement this approach without global influencing. Will try to implement original approach.

alex-zaitsev:
Fixed in https://github.com/Altinity/clickhouse-operator/pull/1623 by .spec.suspend attribute

PR #1580, opened 2024-12-03
feat: ability to suspend `ClickHouseInstallation` and `ClickHouseKeeper` reconciliation

a7i:
For ClickHouse technical maintenance or debugging, it would be ideal to suspend reconciliation. For our use-case, we plan on suspending CRs during Blue/Green cluster upgrades, so as we move workloads from one cluster to another (one namespace at a time), we would need the operator to suspend any reconciliation and only resume once the namespace is fully migrated to the new cluster

Closes https://github.com/Altinity/clickhouse-operator/issues/1433

Thanks for taking the time to contribute to `clickhouse-operator`!

Please, read carefully [instructions on how to make a Pull Request](https://github.com/Altinity/clickhouse-operator/blob/master/CONTRIBUTING.md#intro).

This will help a lot for maintainers to adopt your Pull Request. 

--

<sup>1</sup> If you feel your PR does not affect any Go-code or any testable functionality (for example, PR contains docs only or supplementary materials), PR can be made into `master` branch, but it has to be confirmed by project's maintainer.

### DIFF
--- a/deploy/builder/templates-install-bundle/clickhouse-operator-install-yaml-template-01-section-crd-01-chi-chit.yaml
+++ b/deploy/builder/templates-install-bundle/clickhouse-operator-install-yaml-template-01-section-crd-01-chi-chit.yaml
@@ -92,6 +92,11 @@ spec:
           description: Age of the resource
           # Displayed in all priorities
           jsonPath: .metadata.creationTimestamp
+        - name: suspend
+          type: string
+          description: Suspend reconciliation
+          # Displayed in all priorities
+          jsonPath: .spec.suspend
       subresources:
         status: {}
       schema:
@@ -308,6 +313,13 @@ spec:
                   enum:
                     - ""
                     - "RollingUpdate"
+                suspend:
+                  <<: *TypeStringBool
+                  description: |
+                    Suspend reconciliation of resources managed by a ClickHouse Installation.
+                    Works as the following:
+                     - When `suspend` is `true` operator stops reconciling all resources.
+                     - When `suspend` is `false` or not set, operator reconciles all resources.
                 troubleshoot:
                   <<: *TypeStringBool
                   description: |
--- a/deploy/builder/templates-install-bundle/clickhouse-operator-install-yaml-template-01-section-crd-03-chk.yaml
+++ b/deploy/builder/templates-install-bundle/clickhouse-operator-install-yaml-template-01-section-crd-03-chk.yaml
@@ -88,6 +88,11 @@ spec:
           description: Age of the resource
           # Displayed in all priorities
           jsonPath: .metadata.creationTimestamp
+        - name: suspend
+          type: string
+          description: Suspend reconciliation
+          # Displayed in all priorities
+          jsonPath: .spec.suspend
       subresources:
         status: {}
       schema:
@@ -296,6 +301,13 @@ spec:
                     - "disabled"
                     - "Enabled"
                     - "enabled"
+                suspend:
+                  <<: *TypeStringBool
+                  description: |
+                    Suspend reconciliation of resources managed by a ClickHouse Keeper.
+                    Works as the following:
+                     - When `suspend` is `true` operator stops reconciling all keeper resources.
+                     - When `suspend` is `false` or not set, operator reconciles all keeper resources.
                 namespaceDomainPattern:
                   type: string
                   description: |
--- a/pkg/apis/clickhouse-keeper.altinity.com/v1/type_spec.go
+++ b/pkg/apis/clickhouse-keeper.altinity.com/v1/type_spec.go
@@ -23,6 +23,7 @@ import (
 type ChkSpec struct {
 	TaskID                 *types.String       `json:"taskID,omitempty"                 yaml:"taskID,omitempty"`
 	NamespaceDomainPattern *types.String       `json:"namespaceDomainPattern,omitempty" yaml:"namespaceDomainPattern,omitempty"`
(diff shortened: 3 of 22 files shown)


sunsingerus (changes requested):
new field needs to be specified not only in go files, but in CRD spec as well
check deploy/builder

a7i:
> new field needs to be specified not only in go files, but in CRD spec as well check deploy/builder

hey @sunsingerus can you be a bit more specific? the docs are non-existent and I spent a long time running every command in `/dev` with no avail

sunsingerus:
New fields provided by the CRD has to be specified in CRD definition yaml file, such as:
```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: clickhousekeeperinstallations.clickhouse-keeper.altinity.com
```
Templates for these files for CHI and CHK can support tools be found in `deploy/builder` folder

deploy/builder/templates-install-bundle/clickhouse-operator-install-yaml-template-01-section-crd-01-chi-chit.yaml
```diff
@@ -308,6 +313,38 @@ spec:
                   enum:
                     - ""
                     - "RollingUpdate"
+                suspend: &TypeStringBool
```
sunsingerus, line 316: This is duplicate of the definition of the new type. First definition is made for 
```yaml
                stop: &TypeStringBool
                  type: string
```
please change to usage of the type, such as for example for
```yaml
                troubleshoot:
                  <<: *TypeStringBool
                  description: |
                    Allows to troubleshoot Pods during CrashLoopBack state.
```

sunsingerus (changes requested):

deploy/builder/templates-install-bundle/clickhouse-operator-install-yaml-template-01-section-crd-01-chi-chit.yaml
a7i: ideally you use something like kubebuilder to generate these things

Example: https://github.com/kubernetes-sigs/mcs-api/blob/master/pkg/apis/v1alpha1/serviceexport.go#L24

deploy/builder/templates-install-bundle/clickhouse-operator-install-yaml-template-01-section-crd-01-chi-chit.yaml
```diff
@@ -308,6 +313,13 @@ spec:
...
```
sunsingerus, line 316: Syntax is incorrect still, good example is `troubleshoot` option

deploy/builder/templates-install-bundle/clickhouse-operator-install-yaml-template-01-section-crd-03-chk.yaml
```diff
@@ -296,6 +301,13 @@ spec:
                     - "disabled"
                     - "Enabled"
                     - "enabled"
+                suspend: &TypeStringBool
```
sunsingerus, line 304: Syntax is incorrect still, good example is `troubleshoot` option

sunsingerus (changes requested):

deploy/builder/templates-install-bundle/clickhouse-operator-install-yaml-template-01-section-crd-01-chi-chit.yaml
a7i: Screenshot 2024-12-09 at 10 28 27 AM

pkg/apis/clickhouse-keeper.altinity.com/v1/type_spec.go
```diff
@@ -23,6 +23,7 @@ import (
 type ChkSpec struct {
 	TaskID                 *types.String       `json:"taskID,omitempty"                 yaml:"taskID,omitempty"`
 	NamespaceDomainPattern *types.String       `json:"namespaceDomainPattern,omitempty" yaml:"namespaceDomainPattern,omitempty"`
+	Suspend                *types.StringBool   `json:"suspend,omitempty"                yaml:"suspend,omitempty"`
```
dmvolod, line 26: For pause/suspend option it would be nice to use annotations, like Cluster API project does
Pros: 
- No real object cast needed in predicate functions, just ObjectMeta extraction
- Common approach
- Transparent field migration

sunsingerus: Annotation-based implementation may be added as a second approach, both ways are possible and have pros/cons

a7i: I am generally not a big fan of "annotations as an API", due to lack of schema validation and consistency.

There are many GitHub issues where kubernetes maintainers are also aligned on this. As mentioned above, there are pros/cons to each, but if you can drive it via a defined API, that's always best.

>https://github.com/kubernetes/kubernetes/issues/101776#issuecomment-841441066

"we should avoid using annotations as an API wherever possible"

>https://github.com/kubernetes/enhancements/issues/2255#issuecomment-1382598613

"I'm not in love with annotations as APIs. Do we REALLY think this is the best answer?"

dmvolod: Ok, make sense, but sometimes annotations a more convenient option, when your work with unstructured or schemaless objects.
Agree with @sunsingerus to have both annotation and field based management is a good way.

sunsingerus (approved):

PR #1585, opened 2024-12-06
fix: release_chart workflow bugs

### LINKED ISSUE #1584: Helm Chart release failed for 0.24.1 and 0.24.2
See https://github.com/Altinity/clickhouse-operator/actions/workflows/release_chart.yaml

janeklb:
Was missing an `id` on a step that was referenced by a latter step, and also missing `}}` closing parens

fixes #1584


Slach:
thanks for catch, will backport to 0.24.1 and 0.24.2

PR #1640, opened 2025-02-17
feat(helm): add namespace override capability for multi-namespace deployments

bwdmr:
Adds a new helper template "altinity-clickhouse-operator.namespace" that allows overriding the release namespace via Values.namespaceOverride. Updates all resource templates to use this helper instead of directly referencing .Release.Namespace. This change enables multi-namespace deployments in combined charts while maintaining backward compatibility.

### DIFF
--- a/deploy/helm/clickhouse-operator/values.yaml
+++ b/deploy/helm/clickhouse-operator/values.yaml
@@ -1,3 +1,5 @@
+namespaceOverride: ""
+
 operator:
   image:
     # operator.image.repository -- image repository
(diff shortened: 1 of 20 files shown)


Slach (changes requested):
look to `dev/generate_helm_chart.sh`

all templates/generated shall be changed from this script

deploy/helm/clickhouse-operator/values.yaml
```diff
@@ -755,21 +755,21 @@ additionalResources: []
 #    kind: ConfigMap
 #    metadata:
 #      name: {{ include "altinity-clickhouse-operator.fullname" . }}-cm
-#      namespace: {{ .Release.Namespace }}
```
Slach, line 758: Actually, I don't understand, what this kind of changes actually means?

Where did you define `altinity-cloud-opeator.namespace`?

Slach (changes requested):

Slach:
Sorry for disturbing you again

Could you clarify?
What is the goal for these changes?

Do you want to install helm chart to some namespace but install rendered manifest to another namespace?
Why?

Which behavior you try to achieve? 

If you want to change operator watched namespace for example and install operator in one namespace but watch custom resources in another namespaces

just change `values.yaml` 
```yaml
  files:
    config.yaml:
      # empty value will watch all namespaces
      namespaces: [""]
```

Slach (approved):

Slach:
@bwdmr thanks for contribution

PR #1657, opened 2025-03-13
fixing a typo in keeper prometheus alert

morkalfon:
Fixing a mistyped name in the Keeper alerts that ships with the default rules pack.


Slach (approved):

Slach:
thanks for contribution

PR #1671, opened 2025-04-03
Fix rootCA verification

realyota:
Fix operator rootCA verification by properly decoding pem certificate before ParseCertificate func call


alex-zaitsev:
@realyota , could you add a test manifest to check this? No need to write the test itself, but just provide a test CHI. I will wrap it to a test later. The latest one is test-055, so create test-056-x509.yaml. Ideally, we want to map via a secret file, so it could be reloaded, but I can do it later if this is a problem.

realyota:
@alex-zaitsev @Slach added test manifests.

alex-zaitsev:
Let's make a fallback to the previous code. It may happen that users are already using decoded certificates.

realyota:
maybe just add check, if rootCA starts with PEM header then use decode, if not stay with old behaviour.

realyota:
@alex-zaitsev on the second thought

>  It may happen that users are already using decoded certificates.

This is not possible, since certificate is registered later using `RegisterTLSConfig` without encoding it back (it expect x509 format). So the code itself is invalid. Please review.

```
                         rootCAs.AddCert(cert)
			if err := goch.RegisterTLSConfig(tlsSettings, &tls.Config{
				RootCAs: rootCAs,
			}); err != nil {
				c.l.V(1).F().Error("unable to register TLS config %v", err)
			}
```

sunsingerus:
merged manually

PR #1732, opened 2025-06-02
add example for schema bootstrap via startup_scripts, fix https://git…

Slach:
add example for schema bootstrap via startup_scripts, fix https://github.com/Altinity/clickhouse-operator/issues/1730


sunsingerus (changes requested):
Do not think this example should be in `02-templates-` section, cause it has no additional value in terms of templates usage demonstration. May be we should start a new section, something like `Advanced tuning` or whatever where examples such as UDF and this startup scripts should go.

Slach:
> Do not think this example should be in `02-templates-` section, cause it has no additional value in terms of templates usage demonstration. May be we should start a new section, something like `Advanced tuning` or whatever where examples such as UDF and this startup scripts should go.

ok. example renamed in https://github.com/Altinity/clickhouse-operator/pull/1732/commits/5cfd434c09fc3521f083364f1301343436a3a5f2

sunsingerus (approved):

PR #1739, opened 2025-06-06
feat: users with SSL Auth

### LINKED ISSUE #1727: User with ssl_certificates auth set in CHI generated wrong
Operator version: 0.25.0

I'm trying to create user over CHI which will use SSL certificate for authentication 

```
    users:
      some_user/ssl_certificates/common_name: clickhouse-client-cert
      some_user/grants/query:
        - "GRANT dwh_reader_role"
        - "GRANT SELECT, INSERT ON test_db.some_table"
      some_user/networks/ip:
        - 10.20.0.0/16
```

Converts into in ConfigMap (common-userd):
```
            <some_user>
                <grants>
                    <query>GRANT dwh_reader_role</query>
                    <query>SELECT, INSERT ON test_db.some_table</query>
                </grants>
                <networks>
                    <host_regexp>(chi-clickhouse-[^.]+\d+-\d+|clickhouse\-clickhouse)\.clickhouse\.svc\.cluster\.local$</host_regexp>
                    <ip>::1</ip>
                    <ip>127.0.0.1</ip>
                    <ip>10.20.0.0/16</ip>
                </networks>
              <password_sha256_hex>37a8eec1ce19687d132fe29051dca629d164e2c4958ba141d5f4133a33f0688f</password_sha256_hex>
                <profile>default</profile>
                <quota>default</quota>
                <ssl_certificates>
                    <common_name>clickhouse-client-cert</common_name>
                </ssl_certificates>
            </some_user>
```

And this leads to error:

> <Error> Application: Caught exception while setting up access control.: Code: 347. DB::Exception: Code: 36. DB::Exception: More than one field of 'password', 'password_sha256_hex', 'password_double_sha1_hex', 'no_password', 'ldap', 'kerberos', 'ssl_certificates', 'ssh_keys', 'http_authentication' are used to specify authentication info for user some_user. Must be only one of them.: while parsing user 'some_user' in users configuration file: while loading configuration file '/etc/clickhouse-server/users.xml'. (BAD_ARGUMENTS), Stack trace (when copying this message, always include the lines below)

Generated user in xml have password but shouldn't. How can I prevent generation password for user, which uses ssl_certificates instead of password?

viktordebulat:
This PR makes possible to add [user with SSL authentication](https://clickhouse.com/docs/guides/sre/ssl-user-auth), fix #1727 

Current logic in user normalizer prevents doing that because if password not set, operator adds it during reconciliation and this brakes CH during startup due to invalid config (either password should be set, or ssl_certificate, not both)

### DIFF
--- a/docs/chi-examples/22-secure-ssl-04-users-ssl-auth.yaml
+++ b/docs/chi-examples/22-secure-ssl-04-users-ssl-auth.yaml
@@ -1,0 +1,208 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: clickhouse-cert
+type: Opaque
+stringData:
+  ca.crt: |
+    -----BEGIN CERTIFICATE-----
+    MIIDETCCAfmgAwIBAgIUU3gzeBiW1NUIwfI3WXP+5Mqt9d4wDQYJKoZIhvcNAQEL
+    BQAwGDEWMBQGA1UEAwwNQ2xpY2tIb3VzZSBDQTAeFw0yNTA2MDYxMjAwMjhaFw0z
+    NTA2MDQxMjAwMjhaMBgxFjAUBgNVBAMMDUNsaWNrSG91c2UgQ0EwggEiMA0GCSqG
+    SIb3DQEBAQUAA4IBDwAwggEKAoIBAQDJTN0iu/lvMwTDdygT+h/yXXjZwYJB1aDR
+    oItEWUoc1KfcKFkzk2I3hAbUWzcT3ErhmS1KNBc015MvXTnbTgpItt5xA2lmD39K
+    6bEG8SZ6MFs/RLR2TxYfQw8SNB87//OBHmNrgWK3l4ugFLS8JknZSufxgKUcHaeR
+    2twByCZREPPJFCfo2SAqytBlAHNd9llX7dPj9xgMd1Y7N+hvRF0OJRI5OOjSQtaU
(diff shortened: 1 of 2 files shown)


docs/chi-examples/22-secure-ssl-03-files-multi-secrets-ref.yaml
```diff
@@ -108,13 +108,16 @@ spec:
     users:
       user1/password: qwerty
       user1/networks/ip: "::/0"
+      # Or use SSL Auth (use own CN from certificate), without password
```
Slach, line 111: maybe better move these changes into separate file where we provide `ssl_certificates` example?

viktordebulat: Good point. Will it be ok to use cert-provider CRDs for it not to mess with those secrets? It is used almost in every environment nowadays

Slach (approved):

docs/chi-examples/22-secure-ssl-03-files-multi-secrets-ref.yaml
Slach: no, cert-provider here will over engineering and will too complicated example for understanding

viktordebulat: Pushed with separate example @Slach

Issue #1706, opened 2025-05-13
PodDisruptionBudget "maxUnavailable" doesn't prevent from downtime


varnastadeus:
### Description

Currently, `PodDisruptionBudget` allows configuring only `maxUnavailable`, which doesn't prevent downtime while evicting shard pods. This is reproducible with a simple example having a single shard and 2 replicas and default `maxUnavailable: 1`.

```
     layout:
        shards:
        - internalReplication: "true"
          replicas:
          - templates:
              podTemplate: clickhouse-in-zone-a
          - templates:
              podTemplate: clickhouse-in-zone-b
```

While evicting both pods at a time, there is a short gap when 2nd pod can be evicted while the first pod is gone. 

The status of `pdb` before start of the eviction:

```
{
  "conditions": [
    {
      "lastTransitionTime": "2025-05-11T16:05:17Z",
      "message": "",
      "observedGeneration": 1,
      "reason": "SufficientPods",
      "status": "True",
      "type": "DisruptionAllowed"
    }
  ],
  "currentHealthy": 2, // <- main part
  "desiredHealthy": 1, // <- main part
  "disruptionsAllowed": 1, // <- main part
  "expectedPods": 2,
  "observedGeneration": 1
}
```
The status of `pdb` when 1st pod was deleted:
```
{
  "conditions": [
    {
      "lastTransitionTime": "2025-05-11T16:05:58Z",
      "message": "",
      "observedGeneration": 1,
      "reason": "SufficientPods",
      "status": "True",
      "type": "DisruptionAllowed"
    }
  ],
  "currentHealthy": 1, // <- main part
  "desiredHealthy": 0, // <- main part
  "disruptionsAllowed": 1, // <- main part
  "expectedPods": 1,
  "observedGeneration": 1
}
```
Notice that at some point, pdb observes`"expectedPods": 1`. While it happens, 2nd pod is allowed to be evicted as it doesn't violate the pod's disruption budget based on the current state.

Based on k8s docs, we can't be using `maxUnavailable` with the pod setup we have:

https://kubernetes.io/docs/tasks/run-application/configure-pdb/#arbitrary-controllers-and-selectors
```
You can use a PDB with pods controlled by another resource, by an "operator", or bare pods, but with these restrictions:

- only .spec.minAvailable can be used, not .spec.maxUnavailable.
- only an integer value can be used with .spec.minAvailable, not a percentage.
```

[...]

PR #1768, opened 2025-07-10
Add pdbManaged for CH installation

zrudzionis:
### Description
Allow disabling PDB management for CH installation.

Should help resolve:
https://github.com/Altinity/clickhouse-operator/issues/1706


alex-zaitsev:
Thank you @zrudzionis . We have planned to do similar for 0.25.3 so you saved our time.

PR #1778, opened 2025-07-22
Fix mistype

zrudzionis:
### Description
Fix mistype


Slach:
nice catch, weird why it not crash tests earlier ;(

zrudzionis:
> nice catch, weird why it not crash tests earlier ;(

We don't have a test that is setting `max_unavailable` via dict, only as plain integer. New format:
```

            "pdb": {"single": {"is_managed": False}},
```

old format:
```
           "pdb": {"simple": 1},
```

PR #1838, opened 2025-10-05
fix: add chi-/chk- prefixes to PDB names to prevent naming conflicts

karnzx:
## Problem

When both ClickHouse Installation (CHI) and ClickHouse Keeper (CHK) resources are deployed in the same namespace with the same name, their PodDisruptionBudgets (PDBs) will have identical names, causing one PDB to be replaced by the other. This results in only one PDB being created instead of two separate PDBs for each resource type.

### Root Cause

The current implementation creates PDB names using only the custom resource name without any prefix:

```go
  ObjectMeta: meta.ObjectMeta{
   Name:            c.nm.Name(interfaces.NameClusterPDB, cluster),
   ...
  },
```

## Solution

Add the `chi-` and `chk-` prefix to custom resource PDB names to match the existing pattern used in PDB creation and ensure unique naming across resource types.

## Impact Assessment

- **Very Low Risk**: This is a non-breaking change that only affects PDB naming
- **No Breaking Changes**: Existing deployments will continue to work; only new PDBs will have the prefix
- **Backward Compatibility**: Existing PDBs with and without prefixes will continue to function normally
- **Consistency**: Aligns PDB naming with the existing naming pattern

## Example

**Before the fix:**

- CHI resource named "my-cluster" → PDB name: "my-cluster"
- CHK resource named "my-cluster" → PDB name: "my-cluster" (conflict!)

**After the fix:**

- CHI resource named "my-cluster" → PDB name: "chi-my-cluster-<cluster-name>"
- CHK resource named "my-cluster" → PDB name: "chk-my-cluster" (no conflict!)

This ensures both PDBs can coexist in the same namespace without naming conflicts.

### DIFF
--- a/pkg/model/common/creator/pdb.go
+++ b/pkg/model/common/creator/pdb.go
@@ -15,23 +15,37 @@
 package creator
 
 import (
+	"fmt"
+
 	policy "k8s.io/api/policy/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
+	apiChk "github.com/altinity/clickhouse-operator/pkg/apis/clickhouse-keeper.altinity.com/v1"
 	api "github.com/altinity/clickhouse-operator/pkg/apis/clickhouse.altinity.com/v1"
 	"github.com/altinity/clickhouse-operator/pkg/interfaces"
 )
 
 // CreatePodDisruptionBudget creates new PodDisruptionBudget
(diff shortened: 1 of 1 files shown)


alex-zaitsev:
@karnzx , that has been fixed in 0.25.3. Are you using the latest version? The naming patterns for PDBs is the following now:
[chk|chi]-{name}-{cluster}

sunsingerus:
fixed in [release 0.25.3](https://github.com/Altinity/clickhouse-operator/releases/tag/release-0.25.3)

Issue #1889, opened 2025-12-12
Make OCI charts published to GHCR public


DrFaust92:
OCI charts are pushed here https://github.com/Altinity/clickhouse-operator/blob/master/.github/workflows/release_chart.yaml#L86

it would be great to consume these as OCI for modern workflows. this a change on the repo settings i believe.

ref for similar request: https://github.com/robusta-dev/robusta/pull/1954#issuecomment-3638113629

Slach:
should be fixed in https://github.com/Altinity/clickhouse-operator/pull/1891

try 
`helm install oci://ghcr.io/altinity/clickhouse-operator-helm-chart/altinity-clickhouse-operator`

DrFaust92:
[Slach](https://github.com/Slach) It does! thanks a lot!

PR #1891, opened 2025-12-13
hotfix for helm release after failure github actions 

Slach:
look https://github.com/Altinity/clickhouse-operator/actions/runs/20135191099/job/57967523183

should also fix https://github.com/Altinity/clickhouse-operator/issues/1889

### DIFF
--- a/.github/workflows/release_chart.yaml
+++ b/.github/workflows/release_chart.yaml
@@ -5,6 +5,7 @@ on:
     types:
     - published
     - edited
+  workflow_dispatch: {}
 
 jobs:
   release_chart:


.github/workflows/release_chart.yaml
```diff
@@ -20,14 +21,15 @@ jobs:
 
       - name: Install chart-releaser
         run: |
-          wget https://github.com/helm/chart-releaser/releases/download/v1.4.1/chart-releaser_1.4.1_linux_amd64.tar.gz
-          tar -zxf chart-releaser_1.4.1_linux_amd64.tar.gz cr
+          HELM_CHART_RELEASER_VERSION=$(curl -fsSL https://github.com/helm/chart-releaser/releases/latest -H "Accept: application/json" | jq -r .tag_name)
```
DrFaust92: how about https://github.com/helm/chart-releaser-action?

https://github.com/runatlantis/helm-charts/blob/main/.github/workflows/release.yaml  i think it will simplify flow

PR #1928, opened 2026-02-19
helm: add optional oci_registry for images

lesandie:
# PR Summary

Adds optional oci_registry prefixes for Helm chart images so users can inject environment‑specific registries (e.g., ECR) without hardcoding full image paths. The operator and metrics images now render as <oci_registry>/<repository>:<tag> when set, and unchanged when not set, with safe / handling. Documentation updated to clarify correct usage and avoid double‑prefixing.

```yaml
# values.yaml
operator:
  image:
    oci_registry: "1234567890.dkr.ecr.us-east-1.amazonaws.com"
    repository: "altinity/clickhouse-operator"
    tag: "0.26.0"

metrics:
  image:
    oci_registry: "1234567890.dkr.ecr.us-east-1.amazonaws.com"
    repository: "altinity/metrics-exporter"
    tag: "0.26.0"
```

### DIFF
--- a/deploy/helm/clickhouse-operator/README.md
+++ b/deploy/helm/clickhouse-operator/README.md
@@ -95,6 +95,7 @@ crdHook:
 | metrics.enabled | bool | `true` |  |
 | metrics.env | list | `[]` | additional environment variables for the deployment of metrics-exporter containers possible format value `[{"name": "SAMPLE", "value": "text"}]` |
 | metrics.image.pullPolicy | string | `"IfNotPresent"` | image pull policy |
+| metrics.image.registry | string | `""` | optional image registry prefix (e.g. 1234567890.dkr.ecr.us-east-1.amazonaws.com) |
 | metrics.image.repository | string | `"altinity/metrics-exporter"` | image repository |
 | metrics.image.tag | string | `""` | image tag (chart's appVersion value will be used if not set) |
 | metrics.resources | object | `{}` | custom resource configuration |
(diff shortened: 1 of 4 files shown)


lesandie:
@gcodispot This is related to the feature request we talked about.

lesandie:
The registry prefix is now only prepended when repository does not already include a registry host. This avoids double‑prefixing even if users set oci_registry and use a fully qualified repository.

sunsingerus (changes requested):
Thanks for the contribution — the use case is valid and the approach is mostly right. Two changes needed before merging:

## 1. Rename `oci_registry` → `registry`

The standard Helm convention (used by Bitnami and most major charts) is `registry`, not `oci_registry`. Underscores are also inconsistent with the rest of the chart's camelCase naming (`pullPolicy`, `imagePullSecrets`, etc.).

```yaml
operator:
  image:
    registry: ""          # was: oci_registry
    repository: "altinity/clickhouse-operator"
```

## 2. Simplify the template — drop the regex

The double-prefix protection regex adds unnecessary complexity and is fragile. If a user sets `registry`, it is their responsibility to ensure `repository` is a relative path. The template can be reduced to:

```
{{ if .Values.operator.image.registry }}{{ .Values.operator.image.registry | trimSuffix "/" }}/{{ end }}{{ .Values.operator.image.repository }}:{{ include "altinity-clickhouse-operator.operator.tag" . }}
```

Same for the metrics image.

---

Optional (not blocking): consider adding a `global.imageRegistry` that both operator and metrics images fall back to, so users don't have to repeat the registry in two places. But that can be a follow-up PR.

Slach:
guys, why just not use

```yaml
operator:
  image:
    repository: "1234567890.dkr.ecr.us-east-1.amazonaws.com/altinity/clickhouse-operator"
    tag: "0.26.0"

metrics:
  image:
    repository: "1234567890.dkr.ecr.us-east-1.amazonaws.com/altinity/metrics-exporter"
    tag: "0.26.0"
```

lesandie:
@sunsingerus I've rolledback the regexp (I started doing it as you proposed but complicated it 😄 ) and applied the rest of recommendations. WDYT about @Slach suggestion? It is simpler TBH

lesandie:
@sunsingerus We can continue with the current implementation. BTW do you think it is good if I add the `global.imageRegistry` fall back in this PR?

Issue #1874, opened 2025-11-27
Large Cluster Configmap Limits


Jhors2:
Hi All,
I recently hit an issue where I was trying to provision a large Clickhouse cluster (1000 shards) and hit a limit on the amount of nodes that could exist in the relevant configmap.  After 807 nodes are provisioned the configmap is larger than 1MB in size which is a hard limit in K8S/etcd.  I am wondering if there is a workaround or ideas on how to partition the config maps appropriately to work around this issue.

shankeleven:
I looked into this issue and it seems the failure occurs because the `remote_servers` topology for very large CHIs ends up exceeding Kubernetes' ConfigMap size limits (~1 MiB enforced by the API server/etcd).

The idea is to split the generated `remote_servers` configuration into deterministic fragments instead of storing the entire topology in a single ConfigMap. Each fragment would correspond to a shard range within a cluster, producing multiple ConfigMaps like:

_`chi-<chi>-<cluster>-remote-servers-shard-<startShard>`_

Each fragment would contain a file such as
`chop-generated-remote_servers-part-<startShard>.xml` mounted into _`/etc/clickhouse-server/config.d/`._

Since [ClickHouse merges all XML files under `config.d`](https://clickhouse.com/docs/operations/configuration-files), this should behave identically from the server’s perspective while keeping each ConfigMap well below the 1 MiB limit.

Fragments would be built per cluster by grouping shards sequentially until a configurable size threshold is reached. The fragment identity would be based on (clusterName, shardStart) so resource names remain deterministic across reconciliations.
We could also add a hash annotation on the StatefulSet template so pods roll when the fragment layout changes.

If this direction sounds reasonable, I can start working on a PR for it.

PR #1948, opened 2026-03-25
Implement remote_servers fragmentation via ConfigMaps

shankeleven:
### Motivation

Large ClickHouse clusters can generate `remote_servers.xml` exceeding Kubernetes ConfigMap size limits (~1 MiB), causing reconcile failures.
Solves #1874 
---

### What this PR does

* Splits `remote_servers.xml` into deterministic fragments when size exceeds threshold
* Adds configuration:

  ```yaml
  clickhouse:
    configuration:
      remoteServers:
        remoteServersSplitThresholdBytes: 614400  # 600 KiB
        maxRemoteServersFragments: 100
  ```
* Preserves legacy behavior when size is below threshold
* Uses projected volume to mount fragments transparently
* Adds hash annotation to trigger deterministic pod rollout
* Implements safe GC:

  * post-StatefulSet reconcile
  * on CHI deletion

---

### Testing

* Unit tests for fragmentation logic and edge cases
* E2E tests for:

  * fragment creation & mounting
  * GC after scale-down
  * max fragment failure scenario

---
### Result
* This removes ConfigMap-size-related reconciliation risk for large deployments while keeping behavior unchanged for clusters that do not require fragmentation.

### DIFF
--- a/config/config-dev.yaml
+++ b/config/config-dev.yaml
@@ -85,6 +85,18 @@ clickhouse:
       # Default host_regexp to limit network connectivity from outside
       hostRegexpTemplate: "(chi-{chi}-[^.]+\\d+-\\d+|clickhouse\\-{chi})\\.{namespace}\\.svc\\.cluster\\.local$"
 
+    ################################################
+    ##
+    ## Configuration remote_servers section
+    ##
+    ################################################
+    remoteServers:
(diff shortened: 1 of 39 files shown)


shankeleven:
@sunsingerus @alex-zaitsev could you please review this PR

PR #1965, opened 2026-04-22
Scope schema discovery to target host's cluster in multi-cluster CHI

### LINKED ISSUE #1964: HostCreateTables uses CHI-wide endpoints for schema discovery, leaks schemas across clusters in multi-cluster CHI
## Summary

When the operator creates tables on a newly added host, the endpoint list used to discover schemas is scoped to the entire `ClickHouseInstallation` rather than to the target host's cluster. In a CHI that defines multiple `clusters`, the operator may pick a host from an unrelated cluster as the query source, which can return an incorrect/incomplete set of CREATE statements for the new node.

The scoping is an old latent design issue (see timeline below); it became reliably reproducible for us after upgrading to `0.26`.

## Details

`HostCreateTables` (`pkg/model/chi/schemer/schemer.go`) calls `getReplicatedObjectsSQLs` / `getDistributedObjectsSQLs`. Both use `api.ClickHouseInstallation{}` as the `Names(NameFQDNs, ...)` scope:

- `pkg/model/chi/schemer/replicated.go` lines 69, 76, 83
- `pkg/model/chi/schemer/distributed.go` lines 59, 66, 73

With that scope, `createFQDNs` in `pkg/model/chi/namer/name.go` routes through `createPodFQDNsOfCHI`, which walks every host in every cluster of the CHI. `QueryUnzip2Columns` then calls `QueryAny` (`pkg/model/clickhouse/cluster.go:76`), which iterates the endpoints sequentially and returns the first one that answers — including hosts that belong to a different cluster than the one being scaled.

The SQL itself does filter by cluster name via `clusterAllReplicas('<target cluster>', system.tables)`, but the replicated-table query in `pkg/model/chi/schemer/sql.go` (around line 205) joins against `system.databases` on the executing node:

```sql
FROM clusterAllReplicas('<cluster>', system.tables) tables
LOCAL JOIN system.databases databases on (databases.name = tables.database)
WHERE database NOT IN (...) AND databases.engine IN ('Ordinary','Atomic','Memory','Lazy')
```

The `system.databases` join resolves against the local databases of whichever node served the query. If `QueryAny` picks a node from a different cluster that does not host the same databases (or hosts them with different engines), schema discovery returns the wrong set of CREATE statements.

Beyond the LOCAL JOIN behavior, the intent is also off: the endpoints used to discover schemas for cluster X should be hosts in cluster X — not any host in the CHI. `shouldCreateReplicatedObjects` already uses `api.ChiShard{}` / `api.Cluster{}` scopes for related logic.

## Why this started showing up after 0.26

Walking the git history on the relevant code:

- **Scoping bug is old.** `CreateFQDNs(host, …ClickHouseInstallation{}…)` for schema discovery dates back to at least 2021 (commit `6b946799d`, "dev: unify CreateFQDNs function"), which itself only renamed an older `CreatePodFQDNsOfCHI(host.GetCHI())` call. Semantically this has always walked all clusters of the CHI.
- **LOCAL JOIN form**: commit `d49187d0b` (2024-06, first included in `release-0.23.6`) switched `sqlCreateTableReplicated` from a `has(subquery, ...)` form to `LOCAL JOIN system.databases`. Both forms filter by the executing node's local databases, but the JOIN form is more strict (inner join) so cross-cluster misses are more visible.
- **DNS rework in 0.26**: commit `6d625de69` (Feb 2026, first included in `release-0.26.0`, "dev: rework k8s DNS") added a trailing dot to `patternNamespaceDomain` (`%s.svc.cluster.local.`) "to force absolute DNS lookup, avoiding slow search-suffix resolution with ndots:5". Before that, cross-namespace/cross-cluster endpoints in the CHI-wide list could fail DNS resolution fast under common k8s DNS configs, causing `QueryAny` to fall through and eventually hit a same-cluster host. After 0.26, every CHI endpoint resolves reliably, and `QueryAny` happily returns from the first endpoint in the slice — which, with CHI-wide scoping, can easily be a host from a different cluster.

So the 0.26 DNS fix did not introduce the scoping bug — it removed an accidental failure-mode that had been masking it.

## Reproduction

Define a CHI with two clusters, e.g. a primary data cluster with multiple replicas and a secondary cluster (a small analytics or monitoring cluster) with a different set of databases / table engines. Create tables only on the primary cluster. Scale the primary cluster by adding a replica. Depending on which endpoint `QueryAny` selects, schema discovery for the new replica may run from the secondary cluster's node and return a result filtered by the secondary cluster's local `system.databases`.

Symptoms:
- Missing or inconsistent tables on the newly added replica relative to existing replicas in the same cluster.
- CREATE statements that correspond to the wrong cluster's schema context.

## Proposed fix

Change `Names(NameFQDNs, host, api.ClickHouseInstallation{}, false)` to `Names(NameFQDNs, host, api.Cluster{}, false)` in both `replicated.go` and `distributed.go`. That scopes schema-discovery endpoints to the target host's cluster, matching the existing intent of these functions and the scoping already used by `shouldCreateReplicatedObjects` / `shouldCreateDistributedObjects`.

PR: #1965

lukas-pfannschmidt-tr:
## Summary

Fixes #1964.

`HostCreateTables` used `api.ClickHouseInstallation{}` as the scope for `Names(NameFQDNs, ...)` when building the endpoint list passed to `QueryUnzip2Columns` / `QueryAny`. In a CHI that defines multiple `clusters`, this walked hosts from every cluster and allowed `QueryAny` to pick a source node outside the target cluster.

The SQL does filter by cluster name via `clusterAllReplicas('<target>', system.tables)`, but `sqlCreateTableReplicated` joins against the executing node's local `system.databases`:

```sql
FROM clusterAllReplicas('<cluster>', system.tables) tables
LOCAL JOIN system.databases databases on (databases.name = tables.database)
WHERE database NOT IN (...) AND databases.engine IN ('Ordinary','Atomic','Memory','Lazy')
```

So the returned set of CREATE statements depends on which cluster's node answered first — resulting in missing or incorrect schemas on newly added replicas in multi-cluster CHIs.

## Timeline (for context)

- The CHI-wide scoping in schema discovery dates back to `CreatePodFQDNsOfCHI(host.GetCHI())` pre-2021 and was preserved through the 2021 unification (`6b946799d`) and the 2024 schemer refactor (`b64a6241d`).
- The `LOCAL JOIN system.databases` form landed in `d49187d0b` (first in `release-0.23.6`), which made the executing-node dependency stricter.
- The 0.26 "rework k8s DNS" commit `6d625de69` added a trailing dot to `patternNamespaceDomain` (`%s.svc.cluster.local.`). That fixed slow/failing DNS resolution under `ndots:5`, but it also removed an accidental failure-mode that had been masking the scoping bug: before 0.26, cross-cluster endpoints in the CHI-wide list could fail DNS quickly and `QueryAny` would fall through to a same-cluster endpoint. After 0.26, every CHI endpoint resolves reliably, so `QueryAny` returns from whichever is first in the slice — which, with CHI-wide scoping, can be a node from a different cluster.

The 0.26 DNS change didn't cause this bug — it exposed it. The root cause is the schema-discovery endpoint scope.

## Change

Use `api.Cluster{}` scope in `getReplicatedObjectsSQLs` and `getDistributedObjectsSQLs`, so schema-discovery endpoints are restricted to the target host's own cluster. This matches the scoping already used by `shouldCreateReplicatedObjects` / `shouldCreateDistributedObjects` for the related gating logic.

Six call sites updated:
- `pkg/model/chi/schemer/replicated.go` (databases, tables, functions)
- `pkg/model/chi/schemer/distributed.go` (databases, tables, functions)

## Test plan

- [x] `go build ./...`
- [ ] Manual verification on a multi-cluster CHI: scale one cluster and confirm new replicas receive the correct schemas sourced from the same cluster.

Single-cluster CHIs are unaffected: cluster-scoped FQDNs equal CHI-scoped FQDNs in that case.

### DIFF
--- a/pkg/model/chi/schemer/distributed.go
+++ b/pkg/model/chi/schemer/distributed.go
@@ -56,21 +56,21 @@ func (s *ClusterSchemer) getDistributedObjectsSQLs(ctx context.Context, host *ap
 	databaseNames, createDatabaseSQLs := debugCreateSQLs(
 		s.QueryUnzip2Columns(
 			ctx,
-			s.Names(interfaces.NameFQDNs, host, api.ClickHouseInstallation{}, false),
+			s.Names(interfaces.NameFQDNs, host, api.Cluster{}, false),
 			s.sqlCreateDatabaseDistributed(host.Runtime.Address.ClusterName),
 		),
 	)
 	tableNames, createTableSQLs := debugCreateSQLs(
 		s.QueryUnzipAndApplyUUIDs(
 			ctx,
-			s.Names(interfaces.NameFQDNs, host, api.ClickHouseInstallation{}, false),
+			s.Names(interfaces.NameFQDNs, host, api.Cluster{}, false),
 			s.sqlCreateTableDistributed(host.Runtime.Address.ClusterName),
 		),
 	)
 	functionNames, createFunctionSQLs := debugCreateSQLs(
 		s.QueryUnzip2Columns(
 			ctx,
-			s.Names(interfaces.NameFQDNs, host, api.ClickHouseInstallation{}, false),
+			s.Names(interfaces.NameFQDNs, host, api.Cluster{}, false),
 			s.sqlCreateFunction(host.Runtime.Address.ClusterName),
 		),
 	)
--- a/pkg/model/chi/schemer/replicated.go
+++ b/pkg/model/chi/schemer/replicated.go
@@ -66,21 +66,21 @@ func (s *ClusterSchemer) getReplicatedObjectsSQLs(ctx context.Context, host *api
 	databaseNames, createDatabaseSQLs := debugCreateSQLs(
 		s.QueryUnzip2Columns(
 			ctx,
-			s.Names(interfaces.NameFQDNs, host, api.ClickHouseInstallation{}, false),
+			s.Names(interfaces.NameFQDNs, host, api.Cluster{}, false),
 			s.sqlCreateDatabaseReplicated(host.Runtime.Address.ClusterName),
 		),
 	)
 	tableNames, createTableSQLs := debugCreateSQLs(
 		s.QueryUnzipAndApplyUUIDs(
 			ctx,
-			s.Names(interfaces.NameFQDNs, host, api.ClickHouseInstallation{}, false),
+			s.Names(interfaces.NameFQDNs, host, api.Cluster{}, false),
 			s.sqlCreateTableReplicated(host.Runtime.Address.ClusterName),
 		),
 	)
 	functionNames, createFunctionSQLs := debugCreateSQLs(
 		s.QueryUnzip2Columns(
 			ctx,
-			s.Names(interfaces.NameFQDNs, host, api.ClickHouseInstallation{}, false),
+			s.Names(interfaces.NameFQDNs, host, api.Cluster{}, false),
 			s.sqlCreateFunction(host.Runtime.Address.ClusterName),
 		),
 	)


sunsingerus (approved):
LGTM after a 3-angle review (correctness / security / code quality).

**What this fixes:** schema discovery in multi-cluster CHIs was using CHI-wide endpoint scope (`api.ClickHouseInstallation{}`) but the surrounding gating (`shouldCreateReplicatedObjects` / `shouldCreateDistributedObjects` at replicated.go:28-29 and distributed.go:28) already used cluster scope. With `LOCAL JOIN system.databases` semantics introduced in d49187d0b and trailing-dot FQDNs in 6d625de69, an off-cluster host could respond to `QueryAny` and surface a different cluster's database catalog — leaking schemas across cluster boundaries inside the same CHI.

**Why it's safe to merge as-is:**
- Six mechanical replacements in two files. Type-switch in `createFQDNs` (pkg/model/chi/namer/name.go:279) already handles the `api.Cluster` scope path.
- Single-cluster CHIs unaffected — cluster-scoped FQDNs equal CHI-scoped FQDNs in that case.
- Aligns discovery scope with the gating that was already cluster-scoped — closes an inconsistency, doesn't introduce one.

**Optional follow-up** (non-blocking): unit test asserting the scope argument passed to `s.Names(interfaces.NameFQDNs, ...)` is `api.Cluster{}` for both `getReplicatedObjectsSQLs` and `getDistributedObjectsSQLs`. The schemer package currently has no tests, so this would establish the pattern.

Issue #1704, opened 2025-05-08
Shard unavailable during reconciliation


cw9:
## Description

We encountered an incident of losing all replicas in a shard during a roll, which led to query failures (or partial data returned when `skip_unavailable_shards` was enabled). Here's a detailed breakdown of the events:

1. A deployment updated the ClickHouse Operator (CHOP) with a config change.
2. CHOP began rolling the CHI: `chi-foo-1`.
3. It successfully rolled `chi-foo-1-0-0-0` and moved on to `chi-foo-1-0-1-0`.
4. `chi-foo-1-0-1-0` was brought down as part of the roll.
5. A Kubernetes node drain event occurred, causing CHOP to be rescheduled to a new node.
6. CHOP re-evaluated the CHI and decided to restart the roll.
7. It began rolling `chi-foo-1-0-0-0` again, even though `chi-foo-1-0-1-0` was still down.
8. As a result, both replicas were temporarily unavailable.
9. During this window, ClickHouse queries either failed or returned partial results due to both replicas being down.
10. Once `chi-foo-1-0-0-0` came back online, queries recovered.
11. CHOP then brought up `chi-foo-1-0-1-0`.

## Expected Behavior

- CHOP should ensure at least one replica remains available during a CHI roll, even if it is rescheduled or interrupted.
- CHOP should try to bring the cluster back to a **steady state** before proceeding with the rollout of a new configuration—specifically, it should bring up all missing replicas from the *previous* state before continuing to roll out the *new* version.

## Environment

- CHOP version: 0.24.2
- ClickHouse version: 25.3

## Questions

We would appreciate guidance on how this situation could be handled more gracefully by CHOP. Specifically:

- Is there a recommended mechanism to ensure the cluster reaches a steady state before continuing with a rollout?
- Could CHOP implement safeguards to prevent simultaneous downtime of all replicas, particularly in scenarios involving operator rescheduling or external interruptions like node drains?
- Are there any existing patterns or features we could leverage to enforce this behavior today?

We’re potentially interested in contributing a fix or enhancement for this issue. If the maintainers agree that improving this behavior makes sense, we’d be happy to discuss possible approaches or take on an implementation with some guidance.

sunsingerus:
Thank you for the detailed incident report — this is a real and well-understood gap in the operator's rolling reconciliation logic. We've confirmed the root cause in the code.

## What is happening

There are two separate gaps that combine to produce this incident:

[...]

PR #1967, opened 2026-04-24
dev: prioritize shard recovery and guard unsafe host disruption

shankeleven:
Solves #1704 

**Problem**
During interrupted rollouts (e.g. operator restart while a replica is down), reconciliation could continue and temporarily take all replicas in a shard offline, causing query failures or partial results.

**Root cause**

- Reconcile was skipped when generation was unchanged, even if some hosts were unhealthy.
- Host processing didn’t prioritize recovery and allowed disruptive actions without checking shard state.

**Changes**

- Health-aware skip: only skip reconcile if generation is unchanged and all hosts are healthy. Otherwise, run recovery.
- Recovery-first ordering: unhealthy hosts are reconciled before rollout.
- Shard safety guard: block/disallow disruptive actions (restart/exclude) if no healthy peer exists in the shard.
- Deferral + retry: unsafe actions are deferred; if nothing progresses, return ErrCRUDAbort to retry via normal controller flow.
- Added unit + regression tests for the above behavior.

 **Validation**

`go test ./pkg/controller/chi/... `passes.

### DIFF
--- a/pkg/controller/chi/controller.go
+++ b/pkg/controller/chi/controller.go
@@ -959,6 +959,15 @@ func (c *Controller) handleObject(obj interface{}) {
 }
 
 func ShouldEnqueue(cr *api.ClickHouseInstallation) bool {
+	if cr == nil {
+		return false
+	}
+
+	if cr.Spec.Suspend.Value() {
+		log.V(2).M(cr).Info("skip enqueue, CHI is suspended")
+		return false
+	}
+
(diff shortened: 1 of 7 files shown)


shankeleven:
@sunsingerus  @alex-zaitsev  please review this PR and let me know if there are any changes required

alex-zaitsev:
Thank you for your contribution, closed in favour of slightly different #2046

Issue #1972, opened 2026-04-29
panic: interface conversion: v1.ICustomResource is nil, not *v1.ClickHouseKeeperInstallation [recovered, repanicked]


mstefany:
clickhouse-operator keeps crashing with following messages:

```
I0429 09:22:13.582532       1 panic.go:860] worker-reconciler-chk.go:58:reconcileCR():end:unknown
2026-04-29T09:22:13Z    INFO    Observed a panic in reconciler: interface conversion: v1.ICustomResource is nil, not *v1.ClickHouseKeeperInstallation   {"controller": "clickhousekeeperinstallation", "controllerGroup": "clickhouse-keeper.altinity.com", "controllerKind": "ClickHouseKeeperInstallation", "ClickHouseKeeperInstallation": {"name":"keeper","namespace":"[REDACTED]"}, "namespace": "[REDACTED]", "name": "keeper", "reconcileID": "[REDACTED]"}
panic: interface conversion: v1.ICustomResource is nil, not *v1.ClickHouseKeeperInstallation [recovered, repanicked]
goroutine 434 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
        /clickhouse-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:115 +0x1fa
panic({0x1d364c0?, 0x41a52a2a4e0?})
        /usr/local/go/src/runtime/panic.go:860 +0x13a
github.com/altinity/clickhouse-operator/pkg/controller/chk.(*Controller).poll(0x41a524042a0, {0x2152860, 0x346ef00}, {0x2177970, 0x41a52776900}, 0x41a52ce3468)
        /clickhouse-operator/pkg/controller/chk/controller.go:146 +0x4dd
github.com/altinity/clickhouse-operator/pkg/controller/chk.(*worker).waitForIPAddresses(0x41a52a0eff0, {0x2152860, 0x346ef00}, 0x41a52776900)
        /clickhouse-operator/pkg/controller/chk/worker-wait-exclude-include-restart.go:46 +0x9b3
github.com/altinity/clickhouse-operator/pkg/controller/chk.(*worker).reconcileCR(0x41a52a0eff0, {0x2152860, 0x346ef00}, 0x0, 0x41a536d8f00)
        /clickhouse-operator/pkg/controller/chk/worker-reconciler-chk.go:125 +0x1fc5
[...]

PR #1973, opened 2026-04-29
fix(chk): prevent nil pointer panic in poll when CR Get fails

wucm667:
Fixes https://github.com/Altinity/clickhouse-operator/issues/1972

## What this PR fixes

The `poll` function in the ClickHouseKeeperInstallation controller panics when `c.kube.CR().Get()` returns an error because `cur` is nil, and the code performs an unchecked type assertion:

```go
cur, err := c.kube.CR().Get(ctx, namespace, name)
if f(cur.(*apiChk.ClickHouseKeeperInstallation), err) {  // ← panics when cur is nil
```

This causes the operator to crashloop with:
```
panic: interface conversion: v1.ICustomResource is nil, not *v1.ClickHouseKeeperInstallation
```

## Changes

- Added a type assertion with ok-check in `pkg/controller/chk/controller.go` `poll()` function
- If the CR cannot be fetched (nil or wrong type), log the error and continue polling instead of panicking
- This follows the same safe pattern already used in `statusUpdateProcess()` in `pkg/controller/chk/kube/cr.go`


wucm667:
Closing — targeting wrong base branch (0.18.2 doesn't have the chk controller). Will reopen on 0.26.3.

Issue #1876, opened 2025-12-01
Option to exclude per-CPU metrics from scraping


orloffv:
### Problem
The ClickHouse Operator collects metrics from ClickHouse nodes including per-CPU system metrics like:

`metric.OSGuestTimeCPU{N}`
`metric.OSIOWaitTimeCPU{N}`
`metric.OSUserTimeCPU{N}`
`metric.CPUFrequencyMHz_{N}`
etc.

On machines with high core counts, this generates an enormous amount of metrics that provide limited value.
### Our Setup

15 shards × 2 replicas = 30 ClickHouse nodes
CPUs with ~**380 cores per node**
Each node exports ~**5,000 CPU metrics**
Total: ~**150,000 CPU metrics** across the cluster

### Impact

/metrics endpoint response time: ~**8 seconds**
/metrics response size: ~**40 MB**
~**95% of all metrics** are these per-CPU metrics

### Why These Metrics Are Less Useful
As [ClickHouse documentation states](https://github.com/ClickHouse/ClickHouse/blob/d3aca4bf10b98d4aece4a7f720dc0e011723ab08/docs/en/operations/system-tables/asynchronous_metrics.md?plain=1#L479C5-L479C19):

> This is a system-wide metric, it includes all the processes on the host machine, not just clickhouse-server.

These metrics reflect the entire host, not ClickHouse specifically, making them less actionable for ClickHouse monitoring.

<img width="1504" height="825" alt="Image" src="https://github.com/user-attachments/assets/55a76415-1c8c-44eb-a0df-f266e09df6ad" />

alex-zaitsev:
Implemented in https://github.com/Altinity/clickhouse-operator/releases/tag/release-0.27.0

PR #1975, opened 2026-04-30
excludeMetricsRegexp -- filter out noisy metrics, e,g CPU and OS

alex-zaitsev:
Added operator config setting to exclude noise metrics from export. By default, CPU and OS metrics are excluded, those are reported by thread and rarely used. 

Closes https://github.com/Altinity/clickhouse-operator/issues/1876

```
    excludeMetricsRegexp:
      - "^metric\\.(OS.*CPU[0-9]+|CPUFrequencyMHz_[0-9]+)$"
```


alex-zaitsev:
Test failure is because 0.27.0 image is pulled instead of PR one. Fixed in
https://github.com/Altinity/clickhouse-operator/commit/e03f659716c27c8c27fec5f69900916b2f675ce9

PR #1980, opened 2026-05-08
fix: use plain errors for clickhouse model messages

immanuwell:
Tiny vet cleanup in the ClickHouse model package.

Go 1.26 flags `fmt.Errorf(s)` when `s` is already a built string. No formatting needed here, so `errors.New(s)` is the straight shot. same behavior, less noise.

Repro before:
```sh
go test ./pkg/model/clickhouse
```

It failed with:
```text
non-constant format string in call to fmt.Errorf
```

After:
```sh
go test ./pkg/model/clickhouse
```

passes

### DIFF
--- a/pkg/model/clickhouse/cluster.go
+++ b/pkg/model/clickhouse/cluster.go
@@ -16,6 +16,7 @@ package clickhouse
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


sunsingerus (approved):
LGTM. Tiny, correct vet cleanup — `fmt.Errorf(s)` → `errors.New(s)` at 3 sites in pkg/model/clickhouse. Verified:

- Functional equivalence: identical `*errors.errorString` output when `s` has no `%` verbs; no callers use `errors.Is`/`errors.As` on these errors.
- Sweep is complete for the package: `go vet ./pkg/model/clickhouse/` flags exactly these 3 sites.
- Go version claim is accurate: go.mod declares `go 1.26.2`.
- Format-injection rationale is real (SQL interpolates user-controlled cluster names), though theoretical in Go (memory-safe fmt).

Approved.

PR #1984, opened 2026-05-14
Update default port names to Istio friendly names

brosenfi:
Update default port names to more Istio friendly names indicating the app protocol.

Note I did not update the "interserver" port name as there appears to be indirectly linked references (within the operator) using that name - I had a local deployment fail when updating this name to "http-interserver".  Also, istio's app protocol auto-detection seems to work fine for this interserver connection.

### DIFF
--- a/pkg/apis/clickhouse.altinity.com/v1/type_host.go
+++ b/pkg/apis/clickhouse.altinity.com/v1/type_host.go
@@ -577,13 +577,13 @@ func (host *Host) HasDesiredStatefulSet() bool {
 }
 
 const (
-	ChDefaultPortName   = "port"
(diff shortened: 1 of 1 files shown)


sunsingerus:
what is the motivation for these changes?

brosenfi:
Hi @sunsingerus - adding the ability to run this operator within istio enabled clusters.  We have noticed without these name changes that istio / envoy is not auto-detecting the protocols successfully in our environment ( see [this description](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/) concerning istio's app protocol selection and app protocol selection as part of the name).

PR #1989, opened 2026-05-22
Fix permanent ArgoCD diff on resourceFieldRef.divisor

pkieszcz:
ArgoCD shows a permanent diff on the operator Deployment because resourceFieldRef.divisor isn't set in the chart, but the apiserver materializes it as a zero-value Quantity that serializes back as '0'.

Issue tracked at Altinity/helm-charts#110.

Fix is to set `divisor: "1"` on all 8 resourceFieldRef blocks (4 per container, 2 containers). "1" is the documented k8s default, so no behavior change - kubelet already uses 1 when divisor is unset.

## How I tested

Edited the 3 source templates in `deploy/builder/`, then ran `deploy/builder/build-clickhouse-operator-install-yaml.sh` and `dev/generate_helm_chart.sh` to regenerate the 7 derived files (install bundles + helm chart Deployment).

- `helm lint` passes
- `helm template` renders 8 `divisor: "1"` lines as expected
- no other resourceFieldRef occurrences in the tree

### DIFF
--- a/deploy/builder/templates-install-bundle/clickhouse-operator-install-yaml-template-04-section-deployment-01-no-configmap.yaml
+++ b/deploy/builder/templates-install-bundle/clickhouse-operator-install-yaml-template-04-section-deployment-01-no-configmap.yaml
@@ -84,21 +84,25 @@ spec:
                 resourceFieldRef:
                   containerName: clickhouse-operator
                   resource: requests.cpu
+                  divisor: "1m"
             - name: OPERATOR_CONTAINER_CPU_LIMIT
               valueFrom:
                 resourceFieldRef:
                   containerName: clickhouse-operator
                   resource: limits.cpu
+                  divisor: "1m"
             - name: OPERATOR_CONTAINER_MEM_REQUEST
               valueFrom:
                 resourceFieldRef:
                   containerName: clickhouse-operator
                   resource: requests.memory
+                  divisor: "1Mi"
             - name: OPERATOR_CONTAINER_MEM_LIMIT
               valueFrom:
                 resourceFieldRef:
                   containerName: clickhouse-operator
                   resource: limits.memory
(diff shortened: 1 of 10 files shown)


Slach:
Thanks for the fix — confirmed it resolves the ArgoCD diff and the operator only logs these env vars (`pkg/chop/config_manager.go:logEnvVarParams`), so there is no runtime impact today.

One request before we merge: could you change the divisors to human-readable units instead of `"1"`?

- For CPU fields (`requests.cpu`, `limits.cpu`) → `divisor: 1m`
- For memory fields (`requests.memory`, `limits.memory`) → `divisor: 1Mi`

Reasoning: with `divisor: "1"`, CPU values get rounded **up** to whole cores (e.g. `cpu: 100m` would be exposed as `"1"`), and memory values are emitted as raw bytes (e.g. `512Mi` → `"536870912"`). That's fine right now because nothing parses these variables, but if anyone ever does, the values will be surprising. `1m` / `1Mi` give millicores and mebibytes respectively, which match how users actually write the limits in the chart.

The apiserver still normalizes these to the same Quantity it currently materializes, so ArgoCD diff should remain resolved. Please update the three source templates under `deploy/builder/` and regenerate.

Slach (changes requested):
https://github.com/Altinity/clickhouse-operator/pull/1989#issuecomment-4544478080

pkieszcz:
Switched all 8 resourceFieldRef blocks to divisor: 1m for cpu and divisor: 1Mi for memory across the three source templates under deploy/builder/, then regenerated the bundles and the helm Deployment.

Verified the round-trip against a live apiserver via kubectl apply --dry-run=server on the rendered Deployment — the divisor field comes back as 1m / 1Mi verbatim, so the ArgoCD diff stays resolved.

Slach (approved):

PR #2007, opened 2026-06-06
fix(helm): wire watchNamespaces value into operator ConfigMap

### LINKED ISSUE #1919: Helm chart v0.25.6: configs.watch.namespaces value is ignored — ConfigMap always renders watch.namespaces: []
## Description

In the Helm chart v0.25.6, setting `configs.watch.namespaces` in user-provided `values.yaml` has **no effect** on the rendered ConfigMap. The `config.yaml` data inside ConfigMap `clickhouse-operator-files` always contains `watch.namespaces: []`, regardless of what is passed via values.

This means the operator only watches its own namespace and never discovers CHI resources in other namespaces — a common deployment pattern (e.g., operator in `clickhouse-operator`, CHI in `clickhouse`).

## Root Cause

The `values.yaml` has **two separate paths** that both define `watch.namespaces`:

1. **`configs.watch.namespaces`** (line ~153 in values.yaml) — appears to be intended as the user-facing config knob, but is **never used** by any template.

2. **`configs.files.config.yaml.watch.namespaces`** (line ~261 in values.yaml) — this is the actual YAML blob embedded as a multi-level dict inside `configs.files`, and it is **hardcoded to `[]`**. The ConfigMap template (`ConfigMap-etc-clickhouse-operator-files.yaml`) renders `configs.files` directly via:
   ```
   data: {{ include "altinity-clickhouse-operator.configmap-data" (list . .Values.configs.files) | nindent 2 }}
   ```

Since the template reads from `configs.files` (which contains the full `config.yaml` structure with `watch.namespaces: []` hardcoded), the top-level `configs.watch.namespaces` value is completely ignored.

## Steps to Reproduce

```bash
# 1. Add the Altinity Helm repo
helm repo add altinity https://helm.altinity.com
helm repo update

# 2. Create a values file that sets watch.namespaces
cat > /tmp/test-values.yaml << 'YAML'
configs:
  watch:
    namespaces:
      - clickhouse
      - my-other-namespace
YAML

# 3. Render the template
helm template test-operator altinity/altinity-clickhouse-operator \
  --version 0.25.6 \
  -n clickhouse-operator \
  -f /tmp/test-values.yaml \
  --show-only templates/generated/ConfigMap-etc-clickhouse-operator-files.yaml \
  2>&1 | grep -A2 'watch'
```

### Expected output
```yaml
    watch:
      namespaces:
        - clickhouse
        - my-other-namespace
```

### Actual output
```yaml
    watch:
      namespaces: []
```

The user-provided `configs.watch.namespaces` is silently ignored.

## Workaround

To actually set watch namespaces, users must override the deeply nested `configs.files` path instead:

```yaml
# This works (but is not discoverable):
configs:
  files:
    config.yaml:
      watch:
        namespaces:
          - clickhouse
```

However, overriding `configs.files.config.yaml` replaces the **entire** `config.yaml` blob due to Helm's deep merge behavior on maps, which means users must copy the full default `config.yaml` content and modify only the watch section — error-prone and not maintainable across upgrades.

An alternative workaround is to patch the ConfigMap after `helm install` and restart the operator pod:

```bash
# Patch ConfigMap and restart
kubectl get configmap clickhouse-operator-files -n clickhouse-operator -o json | \
  python3 -c "
import json, sys, yaml
cm = json.load(sys.stdin)
config = yaml.safe_load(cm['data']['config.yaml'])
config['watch']['namespaces'] = ['clickhouse']
cm['data']['config.yaml'] = yaml.dump(config, default_flow_style=False)
json.dump(cm, sys.stdout)
" | kubectl apply -f -
kubectl delete pod -n clickhouse-operator -l app.kubernetes.io/name=altinity-clickhouse-operator
```

## Suggested Fix

The ConfigMap template (or the `configmap-data` helper) should merge `configs.watch.namespaces` into the `config.yaml` content before rendering, or the `values.yaml` should be restructured so that the user-facing `configs.watch.namespaces` actually controls the value inside `configs.files.config.yaml.watch.namespaces`.

## Environment

- Chart: `altinity/altinity-clickhouse-operator` v0.25.6
- Helm: v3.x / v4.x
- Kubernetes: EKS 1.29–1.33

jtomaszon:
## Problem

Closes #1919.

`configs.files.config.yaml.watch.namespaces.include` is hardcoded to `[]` in `values.yaml` and the ConfigMap template renders it directly — there is no way to override it via a top-level Helm value. Users who need the operator to watch multiple namespaces must manually patch the ConfigMap after install, which is fragile and breaks on upgrades.

## Solution

This PR introduces:

1. A top-level `watchNamespaces` value in `values.yaml` (empty by default — preserves existing behavior):

```yaml
# watchNamespaces -- namespaces where the operator watches for ClickHouseInstallation resources.
# If empty, the operator watches only its own namespace (or all namespaces when running in kube-system).
# Use [".*"] to watch all namespaces.
watchNamespaces: []
```

2. A new `altinity-clickhouse-operator.configmap-files` helper in `_helpers.tpl` that deep-copies `configs.files`, patches `watch.namespaces.include` in-place, and delegates to the existing `configmap-data` helper for rendering.

3. The ConfigMap template updated to call the new helper instead of `configmap-data` directly.

## Usage

```yaml
# values.yaml
watchNamespaces:
  - clickhouse-system
  - my-app-namespace
```

Or via `--set`:

```bash
helm install clickhouse-operator altinity/altinity-clickhouse-operator \
  --set "watchNamespaces={clickhouse-system,my-app-namespace}"
```

## Behavior

- `watchNamespaces: []` (default) — no change, operator watches its own namespace
- `watchNamespaces: [".*"]` — watch all namespaces
- Any non-empty list — operator watches exactly those namespaces

## Testing

```bash
# Default — include stays []
helm template test deploy/helm/clickhouse-operator \
  --show-only templates/generated/ConfigMap-etc-clickhouse-operator-files.yaml | grep -A5 'watch:'

# With namespaces set
helm template test deploy/helm/clickhouse-operator \
  --set 'watchNamespaces={clickhouse-system,everest-dev}' \
  --show-only templates/generated/ConfigMap-etc-clickhouse-operator-files.yaml | grep -A8 'watch:'
```

### DIFF
--- a/deploy/helm/clickhouse-operator/README.md
+++ b/deploy/helm/clickhouse-operator/README.md
@@ -138,4 +138,5 @@ crdHook:
 | serviceMonitor.operatorMetrics.scrapeTimeout | string | `""` |  |
 | tolerations | list | `[]` | tolerations for scheduler pod assignment, check `kubectl explain pod.spec.tolerations` for details |
 | topologySpreadConstraints | list | `[]` |  |
+| watchNamespaces | list | `[]` | namespaces where the operator watches for ClickHouseInstallation resources. If empty, the operator watches only its own namespace (or all namespaces when running in kube-system). Use [".*"] to watch all namespaces. Example: watchNamespaces: ["clickhouse", "my-other-namespace"] |
 
--- a/deploy/helm/clickhouse-operator/templates/_helpers.tpl
+++ b/deploy/helm/clickhouse-operator/templates/_helpers.tpl
@@ -122,3 +122,25 @@ null
 {{- tpl (toYaml (dict $k $v)) $root }}
 {{ end }}
 {{- end }}
+
+{{/*
+altinity-clickhouse-operator.configmap-files merges watchNamespaces into the
+operator config before rendering the ConfigMap data block.
+
+This exists because configs.files.config.yaml.watch.namespaces.include is
+deep inside a nested structure — Helm's values merge cannot target it
+directly. Instead we deepCopy the files map, patch the nested value in-place,
+and pass the result to configmap-data.
+
+Arguments (list): root context, configs.files, watchNamespaces list
+*/}}
+{{- define "altinity-clickhouse-operator.configmap-files" -}}
+{{- $root := index . 0 -}}
+{{- $files := deepCopy (index . 1) -}}
+{{- $watchNamespaces := index . 2 -}}
+{{- if $watchNamespaces -}}
+  {{- $namespaces := index (index (index $files "config.yaml") "watch") "namespaces" -}}
+  {{- $_ := set $namespaces "include" $watchNamespaces -}}
+{{- end -}}
+{{- include "altinity-clickhouse-operator.configmap-data" (list $root $files) -}}
+{{- end -}}
--- a/deploy/helm/clickhouse-operator/templates/generated/ConfigMap-etc-clickhouse-operator-files.yaml
+++ b/deploy/helm/clickhouse-operator/templates/generated/ConfigMap-etc-clickhouse-operator-files.yaml
@@ -11,4 +11,4 @@ metadata:
   namespace: {{ include "altinity-clickhouse-operator.namespace" . }}
   labels: {{ include "altinity-clickhouse-operator.labels" . | nindent 4 }}
   annotations: {{ include "altinity-clickhouse-operator.annotations" . | nindent 4 }}
-data: {{ include "altinity-clickhouse-operator.configmap-data" (list . .Values.configs.files) | nindent 2 }}
+data: {{ include "altinity-clickhouse-operator.configmap-files" (list . .Values.configs.files .Values.watchNamespaces) | nindent 2 }}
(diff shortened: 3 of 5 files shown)


sunsingerus (changes requested):
PR includes modified files that are auto-generated and no modifications to the generator itself

jtomaszon:
Thanks for the review — you're right. I've reworked it so the change comes from the generator, not the generated file.

`dev/generate_helm_chart.sh` now special-cases the `etc-clickhouse-operator-files` ConfigMap in `update_configmap_resource()`, emitting the `configmap-files` helper (which wires `watchNamespaces` into `watch.namespaces.include`) instead of the generic `configmap-data`. Re-running `generate_helm_chart.sh` now reproduces the committed template with no manual edits. The README values table was regenerated by helm-docs.

Verified: default (`watchNamespaces: []`) leaves `include` empty; `--set 'watchNamespaces={ns1,ns2}'` renders them under `watch.namespaces.include`.

Issue #2016, opened 2026-06-30
Feature request: register pprof endpoint for profiling usage


dentiny:
Hi team, we've deployed clickhouse with this clickhouse operator in production.
Sometime we suffered OOM for CH metrics collector sidecar container, so I'm wondering if we could register `pprof` HTTP endpoint, so we could easily profile the process when needed?

- `pprof` is the standard package golang provides: https://pkg.go.dev/net/http/pprof
- There's no runtime performance or resource impact when no profiling request, in other words, it's request-based
- It provides a few useful information to inspect the process, including but not confined to memory, CPU, mutex, etc

I found others also met OOM issue in the past, which I believe also benefits from pprof endpoint: https://github.com/Altinity/clickhouse-operator/issues/176

Thank you!

PR #2017, opened 2026-06-30
Expose /debug/pprof on operator and metrics-exporter HTTP servers

dentiny:
Closes https://github.com/Altinity/clickhouse-operator/issues/2016

Hi team, I'd like to register `pprof` endpoint for profiling purpose, which is useful to understand production issues such as OOM.

### DIFF
--- a/cmd/operator/app/main.go
+++ b/cmd/operator/app/main.go
@@ -24,9 +24,12 @@ import (
 	"syscall"
 
 	log "github.com/altinity/clickhouse-operator/pkg/announcer"
+	metricspprof "github.com/altinity/clickhouse-operator/pkg/metrics/pprof"
 	"github.com/altinity/clickhouse-operator/pkg/util/fips"
 	"github.com/altinity/clickhouse-operator/pkg/version"
 )
+
+const defaultPprofEndpoint = "127.0.0.1:6060"
 
 // CLI parameter variables
 var (
@@ -42,6 +45,12 @@ var (
 	// debugRequest defines request for clickhouse-operator debug run
 	debugRequest bool
 
+	// pprofRequest enables a loopback-only pprof endpoint.
+	pprofRequest bool
+
+	// pprofEndpoint defines pprof listen address.
+	pprofEndpoint string
+
 	// chopConfigFile defines path to clickhouse-operator config file to be used
 	chopConfigFile string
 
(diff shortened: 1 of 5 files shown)


sunsingerus (changes requested):
Thanks for this, @dentiny — enabling profiling to debug OOM (#2016) is genuinely useful and something we want. I have to request changes, though, because as written this exposes `/debug/pprof` in a way that is a security regression for every existing install. Details and a recommended path below.

## Why this can't merge as-is

`import _ "net/http/pprof"` runs the package `init()`, which registers `/debug/pprof/*` on `http.DefaultServeMux`. Both HTTP servers here serve that mux via the `nil` handler:

- `pkg/metrics/operator/machinery.go:110` — operator metrics, default `:9999`, bound `0.0.0.0` (all interfaces) unconditionally.
- `pkg/metrics/clickhouse/rest_server.go:207` — metrics-exporter, default `:8888`, `0.0.0.0` in the default (Plain) IPC mode.

Both ports are published cluster-wide by the `clickhouse-operator-metrics` Service and scraped by Prometheus, and we ship no NetworkPolicy. So after this change **any pod in any namespace** can reach `/debug/pprof` on the operator, unauthenticated:

- **Info disclosure** — `/debug/pprof/heap` and `/allocs` dump live operator memory, which holds ClickHouse credentials, IPC tokens, and Kubernetes API bearer tokens; `/cmdline` leaks all flags; `/goroutine?debug=2` leaks full stacks.
- **DoS** — `/debug/pprof/profile?seconds=N` and `/trace?seconds=N` take an attacker-controlled duration and share the single listener/mux with `/metrics`, degrading the reconcile loop.
- **Bypasses IPC Secure mode** — the token + loopback middleware wraps only the `/chi` handler; pprof is registered at the mux root, outside that protection, even in Secure mode.
- **On by default** — a blank import active at `init()`, with no flag/config/opt-in, so every install silently gains this endpoint on upgrade.

## A "just use a separate mux" fix is not sufficient

Moving pprof to a private `*http.ServeMux` alone does not help: `net/http/pprof`'s `init()` registers on the global `DefaultServeMux` regardless of whether the import is blank or named, so as long as `machinery.go` / `rest_server.go` serve the `nil` handler they keep exposing it. (Verified: a *named* import still leaves `DefaultServeMux` responding `200` on `/debug/pprof/`.)

## Recommended approach for a mergeable version

1. **Don't register on `DefaultServeMux`.** Simplest is to use `runtime/pprof` (no `init()` side-effect) to serve heap/allocs/goroutine — the OOM-relevant profiles — from a private mux. Alternatively keep `net/http/pprof` but convert the `:8888`/`:9999` servers to explicit, non-`nil` muxes so the polluted default mux is never served.
2. **Bind loopback-only** (`127.0.0.1`), never on the metrics ports and never behind the Service — reachable via `kubectl port-forward`.
3. **Default OFF, opt-in via a flag**, consistent with the existing `-debug` / `-fips-info` flags.
4. Consider **omitting `profile`/`trace`** (the CPU/trace DoS vectors) unless separately gated, and log a warning when pprof is enabled so it is visible in `kubectl logs`.

Happy to review a revised version along these lines — thanks again for tackling the OOM-debuggability gap.

dentiny:
Thank you so much for the detailed review! I made a few changes in the updated commit:
- `runtime/pprof` is used instead of `net/http/pprof`, so it does not auto-register anything on `http.DefaultServeMux`
- Forced pprof to bind only to loopback via `validateLoopbackAddress`
- Operator flags is added to decide whether to enable profiling and which port to use
- Explicit private muxes is used instead of global one

sunsingerus:
This PR does not even compile

PR #2041, opened 2026-07-21
Raft-safe ClickHouseKeeperInstallation rescale

filimonov:
Make CHK scale-up/scale-down Raft-safe by replacing the fixed post-create sleep and the all-at-once XML publish with staged, verified membership changes.

Problem: on rescale the operator published a raft_configuration with every desired server at once and only waited a fixed 7s, never confirming committed Raft membership. Two freshly-started Keeper nodes could form a quorum between themselves while the original node still believed it was the whole cluster — a split-brain window. Scale-down deleted StatefulSets/PVCs with no barrier confirming the member had actually left committed membership.

Approach (hardened staged XML; source-verified against Keeper 23.8..master — versioned `reconfig` CAS does not exist in any version, and the XML-diff mode is the reliable control plane):
- Publish new members one at a time, each behind a committed-membership barrier read from the `/keeper/config` znode plus a leader `mntr` synced-followers barrier. Never more than one not-yet-committed server in the published XML (split-brain guard). Emit `<start_as_follower>` for members joining an established cluster.
- Gate CHK `Completed` on committed membership == desired.
- Scale-down: transfer leadership (`rqld`) off departing members, then gate StatefulSet/PVC purge on confirmed committed removal.
- Lease leader election for the CHK controller manager.
- Fail-safe: any barrier timeout / quorum loss requeues without deleting or mutating anything.

Adds a keeper observe client (`pkg/model/chk/keeper`), the controller barriers (`pkg/controller/chk/worker-raft-membership.go`), and e2e coverage including a deterministic split-brain demonstration (RED on the pre-fix operator, GREEN on this branch, across Keeper 23.8/24.8/25.8). 

It contains 6 commits to make review simpler. Check one commit after another.

### DIFF
--- a/config/chk/keeper_config.d/01-keeper-01-default-config.xml
+++ b/config/chk/keeper_config.d/01-keeper-01-default-config.xml
@@ -35,7 +35,7 @@
             stricter list can override this value, but they must keep `ruok` if they
             also use the default operator probes.
         -->
-        <four_letter_word_white_list>conf,cons,crst,envi,ruok,srst,srvr,stat,wchs,dirs,mntr,isro</four_letter_word_white_list>
+        <four_letter_word_white_list>conf,cons,crst,envi,ruok,srst,srvr,stat,wchs,dirs,mntr,isro,rqld</four_letter_word_white_list>
     </keeper_server>
     <listen_host>::</listen_host>
     <listen_host>0.0.0.0</listen_host>
--- a/deploy/operator/clickhouse-operator-install-ansible.yaml
+++ b/deploy/operator/clickhouse-operator-install-ansible.yaml
@@ -6750,7 +6750,7 @@ data:
                 stricter list can override this value, but they must keep `ruok` if they
                 also use the default operator probes.
             -->
-            <four_letter_word_white_list>conf,cons,crst,envi,ruok,srst,srvr,stat,wchs,dirs,mntr,isro</four_letter_word_white_list>
+            <four_letter_word_white_list>conf,cons,crst,envi,ruok,srst,srvr,stat,wchs,dirs,mntr,isro,rqld</four_letter_word_white_list>
         </keeper_server>
         <listen_host>::</listen_host>
         <listen_host>0.0.0.0</listen_host>
--- a/deploy/operator/clickhouse-operator-install-ansible.yaml
+++ b/deploy/operator/clickhouse-operator-install-ansible.yaml
@@ -5497,6 +5497,23 @@ rules:
       - list
 
   #
+  # coordination.k8s.io resources
+  #
+
+  - apiGroups:
+      - coordination.k8s.io
+    resources:
+      - leases
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - patch
+      - delete
+
+  #
   # apps.* resources
   #
 
--- a/deploy/operator/clickhouse-operator-install-bundle-v1beta1.yaml
+++ b/deploy/operator/clickhouse-operator-install-bundle-v1beta1.yaml
@@ -5365,7 +5365,6 @@ metadata:
   namespace: kube-system
   labels:
     clickhouse.altinity.com/chop: 0.27.2
-
 # Template Parameters:
 #
 # NAMESPACE=kube-system
@@ -5451,6 +5450,21 @@ rules:
     verbs:
       - get
       - list
+  #
+  # coordination.k8s.io resources
+  #
+  - apiGroups:
+      - coordination.k8s.io
+    resources:
+      - leases
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - patch
+      - delete
   #
   # apps.* resources
   #
@@ -5618,7 +5632,6 @@ subjects:
   - kind: ServiceAccount
     name: clickhouse-operator
     namespace: kube-system
-
 # Template Parameters:
 #
 # NAMESPACE=kube-system
@@ -5704,6 +5717,21 @@ rules:
     verbs:
       - get
       - list
+  #
+  # coordination.k8s.io resources
+  #
+  - apiGroups:
+      - coordination.k8s.io
+    resources:
+      - leases
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - patch
+      - delete
   #
   # apps.* resources
   #
--- a/deploy/operator/clickhouse-operator-install-bundle-v1beta1.yaml
+++ b/deploy/operator/clickhouse-operator-install-bundle-v1beta1.yaml
@@ -6951,7 +6951,7 @@ data:
                 stricter list can override this value, but they must keep `ruok` if they
                 also use the default operator probes.
             -->
-            <four_letter_word_white_list>conf,cons,crst,envi,ruok,srst,srvr,stat,wchs,dirs,mntr,isro</four_letter_word_white_list>
+            <four_letter_word_white_list>conf,cons,crst,envi,ruok,srst,srvr,stat,wchs,dirs,mntr,isro,rqld</four_letter_word_white_list>
         </keeper_server>
         <listen_host>::</listen_host>
         <listen_host>0.0.0.0</listen_host>
--- a/pkg/controller/chk/worker-raft-membership.go
+++ b/pkg/controller/chk/worker-raft-membership.go
@@ -1,0 +1,366 @@
+// Copyright 2019 Altinity Ltd and/or its affiliates. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package chk
+
+// Raft-safe CHK rescale — design notes (verified against ClickHouse Keeper
+// source, 23.8..master).
+//
+// Control plane: the operator drives membership through the static
+// raft_configuration XML (enable_reconfiguration=false), NOT the client
+// `reconfig` command. A versioned `reconfig` CAS does not exist in any Keeper
+// version (the version arg is rejected unless -1, and no config version is
+// exposed to read), and the client `reconfig` is optimistic/async and not
+// leader-forwarded. The XML-diff path, by contrast, is applied by the leader
+// only, one server at a time, and serialized by NuRaft — the reliable control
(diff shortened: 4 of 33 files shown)


deploy/operator/clickhouse-operator-install-ansible.yaml
```diff
@@ -6705,13 +6722,20 @@ data:
...
+                Keeper versions known to support them. See
+                pkg/model/chk/config/generator.go getCoordinationSettings.
+            -->
             <coordination_settings>
-                <async_replication>1</async_replication>
                 <min_session_timeout_ms>10000</min_session_timeout_ms>
                 <operation_timeout_ms>10000</operation_timeout_ms>
                 <raft_logs_level>information</raft_logs_level>
                 <session_timeout_ms>100000</session_timeout_ms>
-                <use_xid_64>1</use_xid_64>
```
Slach, line 6770: why?

deploy/operator/clickhouse-operator-install-bundle-v1beta1.yaml
```diff
@@ -6895,13 +6923,20 @@ data:
     <clickhouse>
         <asynchronous_metrics_keeper_metrics_only>1</asynchronous_metrics_keeper_metrics_only>
         <keeper_server>
+            <!--
+                async_replication and use_xid_64 are intentionally NOT set here.
+                They are version-gated coordination settings: older ClickHouse Keeper
+                releases reject them with UNKNOWN_SETTING and CrashLoop at startup
```
Slach: i think we will allow support 26.3 keeper for this, old clickhouse-keeper versions actually useless from my point of view without this settings

pkg/controller/chk/worker-raft-membership.go
```diff
@@ -0,0 +1,366 @@
...
+
+// warnRaftGatesSkippedSecureOnly records — loudly — that a Raft
+// rescale-safety gate was skipped for a secure-only Keeper. It always emits a
+// (k8s-deduplicated) Warning event (visible in `kubectl describe`), and — when
+// the operator config has `status.fields.actions` enabled (off by default) —
+// also pushes a Status.Actions entry. The Warning event alone guarantees the
+// degradation is observable regardless of that config flag.
+func (w *worker) warnRaftGatesSkippedSecureOnly(cr api.ICustomResource, gate string) {
+	msg := fmt.Sprintf(
+		"secure-only Keeper: skipping Raft %s — the operator cannot read /keeper/config or 4LW over the closed plaintext port and has no TLS client material to dial the secure port; rescale-safety is NOT verified (pre-branch behavior). See docs/chk-rescale-raft-safety-v3.md.",
```
Slach, line 230: This user-facing Warning event (visible in `kubectl describe`) references `docs/chk-rescale-raft-safety-v3.md`, but the PR doesn't add that file — the link points nowhere. Same dangling reference in code comments in `worker.go`, `interface.go`, `generator.go`, `event-emitter.go`. Please either include the doc in the PR or drop the references (at minimum from the event text).

pkg/controller/chk/worker-raft-membership.go
```diff
@@ -0,0 +1,366 @@
...
+// BEFORE the shrunk raft XML is published: `rqld` on the lowest-ordinal
+// surviving host (23.8-compatible). Best-effort: Keeper can remove a leader
+// on its own (it yields leadership first); the proactive transfer just avoids
+// that riskier path, so failures here only log a warning. The actual safety
+// fence is the purge barrier in clean().
+func (w *worker) prepareRaftScaleDown(ctx context.Context, cr *apiChk.ClickHouseKeeperInstallation) {
+	removed := removedHosts(cr)
+	if len(removed) == 0 {
+		return
+	}
```
Slach, line 283: On a multi-host scale-down (e.g. 3→1) the shrunk XML is published with **all** removals at once, and Keeper's dispatcher serializes them internally (verified against Keeper source: `clusterUpdateWithReconfigDisabledThread` pops one action and waits for commit before the next) — so this is safe. But Keeper logs `LOG_WARNING` "strictly not recommended" whenever the config diff contains more than one change, and the original design removes servers one at a time. The one-at-a-time staging in this PR covers additions only. Consider either staging removals one at a time too, or explicitly documenting the decision to rely on Keeper's internal serialization for removals (and expect the Keeper-side warning in logs).

pkg/controller/chk/worker-raft-membership.go
```diff
@@ -0,0 +1,366 @@
...
+		// don't burn the full waitRaftMembershipConverged timeout every pass.
+		// Fail-safe is preserved (purge stays blocked); on unstop, discovery
+		// re-finds the orphan and this gate runs the barrier again for real.
+		return false
+	}
+	if crIsSecureOnlyKeeper(cr) {
+		// Secure-only Keeper: the barrier cannot connect (plaintext port closed,
+		// no operator TLS client material). Blocking would orphan the STS/PVC
+		// forever; degrade to pre-branch behavior (allow the purge, unverified).
+		w.warnRaftGatesSkippedSecureOnly(cr, "purge barrier")
```
Slach, line 362: The secure-only degradation (skip all barriers, pre-branch behavior) is a pragmatic call and it's honestly surfaced via the Warning event — but it means TLS-only Keeper clusters get none of this PR's safety guarantees. The comment calls real TLS observation "a documented follow-up"; please open a tracking issue so it doesn't get lost.

config/chk/keeper_config.d/01-keeper-01-default-config.xml
```diff
@@ -35,7 +35,7 @@
             stricter list can override this value, but they must keep `ruok` if they
             also use the default operator probes.
         -->
-        <four_letter_word_white_list>conf,cons,crst,envi,ruok,srst,srvr,stat,wchs,dirs,mntr,isro</four_letter_word_white_list>
+        <four_letter_word_white_list>conf,cons,crst,envi,ruok,srst,srvr,stat,wchs,dirs,mntr,isro,rqld</four_letter_word_white_list>
```
Slach, line 38: `rqld` is added to the *default* whitelist only — a user who overrides `four_letter_word_white_list` (the option documented a few lines above) will silently lose proactive leadership transfer on scale-down. That's acceptable since `rqld` is best-effort and the purge barrier is the real fence, but worth a note in the option's doc comment so users know to include `rqld`.

tests/e2e/test_operator.py
```diff
@@ -7776,12 +7825,49 @@ def test_020005(self):
...
+    with Given("Install 3-replica CHK"):
+        kubectl.create_and_check(
+            manifest=chk_manifest_3, kind="chk",
+            check={"pod_count": 3, "do_not_delete": 1},
+        )
+    check_keeper_membership(chk, [0, 1, 2], note="after install")
+
+    with When("Two of three keepers are down (no quorum for removal commit)"):
+        with When("keepers 1 and 2 are deleted"):
+            kubectl.launch(f"delete pod {keeper_pod_name(chk, 1)} {keeper_pod_name(chk, 2)} --wait=false")
```
Slach, line 7855: Deleted pods are immediately recreated by the StatefulSet controller, so quorum can recover within the 60s window — at which point the operator may *legitimately* complete the scale-down and the `sts_count == 3` assertion below fails precisely because the system worked correctly and fast. To make the no-quorum condition hold for the whole observation window, keep quorum broken deterministically instead of a one-shot `delete pod` — e.g. scale the two STSs to 0 replicas (or block raft traffic with a NetworkPolicy) before applying the scale-down manifest, and restore in the cleanup.

pkg/model/chk/config/generator.go
```diff
@@ -88,6 +89,9 @@ func (c *Generator) getRaftConfig(selector *config.HostSelector) string {
 			util.Iline(raft, i, "    <id>%d</id>", getServerId(host))
 			util.Iline(raft, i, "    <hostname>%s</hostname>", c.namer.Name(interfaces.NameInstanceHostname, host))
 			util.Iline(raft, i, "    <port>%d</port>", host.RaftPort.Value())
+			if isJoiningEstablishedCluster(c.cr, host) {
+				util.Iline(raft, i, "    <start_as_follower>true</start_as_follower>")
```
Slach, line 93: Nit: `<start_as_follower>` stays in the published ConfigMap until the next reconcile rewrites the XML (the joiner keeps `ObjectStatusRequested` only during the current pass, but the last-published XML persists). Harmless — the flag only affects boot, and the NuRaft all-followers restriction can't be hit since bootstrap never emits it — but worth a one-line comment so a future reader doesn't treat the lingering flag in the ConfigMap as a bug.

Slach (changes requested):

deploy/operator/clickhouse-operator-install-ansible.yaml
filimonov: version gate was added and reverted (YAGNI)

deploy/operator/clickhouse-operator-install-bundle-v1beta1.yaml
filimonov: :+1:

Slach (approved):

alex-zaitsev:
I suggest we do the following:
1. Make sure we can catch the problem with raft membership with a test 
2. 'start_as_follower' looks like the correct first step to put into default config

Let's see if the problem persists after that.

filimonov:
> **1. Make sure we can catch the problem with a test**

Agreed, but a *deterministic* test needs artificial conditions — synthetic delays or manual network partitioning — because the bug is a race with a self-healing window.

I'll re-add the "unhappy-path" test. It has to break into the cluster with `iptables` to reproduce the exact situation: **the second node hasn't finished syncing with the first at the moment the third one starts.**

> **2. `start_as_follower` looks like the correct first step for the default config**

It's a correct step, but it is **not sufficient on its own.** The problem isn't the *role* we assign to a new member — it's that we don't **track whether Raft actually reacted** to each change before we make the next one.

Raft membership changes are slow. Before touching the quorum again we have to confirm, for each change, that:
1. the config change reached the node,
2. Keeper re-read the config and pushed the change into Raft, and
3. the member was actually added to (or removed from) the committed quorum.

You can't mutate the quorum fire-and-forget.

That's the root cause of the current bug. We publish the new config and start both extra nodes at once, without waiting. If the **third** node joins before the **second** has been fully committed to the quorum, the second and third form their own majority (2 of 3) and elect a leader *without* the original node. And that's **correct, expected Raft behavior** — it's indistinguishable from "only 2 of 3 nodes are online." `start_as_follower` reduces the chance of the fresh nodes starting an election, but it doesn't close the window; only observing each membership change before the next does.

So the key change this PR introduces is **not** `start_as_follower` — that's a minor hardening detail. The core of it is an **observe-wait-confirm mechanism for Raft events**: after every membership change we watch Raft and block until that change is actually committed, before issuing the next one. This is what enforces the correct interleaving and keeps the causal chain intact — each step (config delivered → Keeper applies it → quorum updated) is confirmed to have happened before the next step begins, instead of firing all changes off in parallel and hoping Raft catches up. That serialization is the actual fix; everything else (`start_as_follower`, the scale-down leadership transfer, the `Completed` gate) is built on top of it.

PR #2046, opened 2026-07-24
Protect last healthy shard replica during interrupted rolls (#1704)

alex-zaitsev:
Reconcile unhealthy replicas in the shard prior to healthy ones. Solves #1704. Replaces #1967
## Summary
- Recovery-first shard reconcile with peer-health gate before disruptive updates
- E2E `test_010083` for interrupted broken-image roll + operator restart
## Test plan
- [x] `test_010083`
- [x] smoke: `test_010008*`, `test_0100022`, `test_0100028`

### DIFF
--- a/pkg/controller/chi/worker-reconciler-chi.go
+++ b/pkg/controller/chi/worker-reconciler-chi.go
@@ -52,9 +52,6 @@ func (w *worker) reconcileCR(ctx context.Context, old, new *api.ClickHouseInstal
 	switch {
 	case w.isAfterFinalizerInstalled(old, new):
 		w.a.M(new).F().Info("isAfterFinalizerInstalled - continue reconcile-1")
-	case w.isGenerationTheSame(old, new):
-		log.V(2).M(new).F().Info("isGenerationTheSame() - nothing to do here, exit")
-		return nil
 	}
 
 	w.a.M(new).S().P()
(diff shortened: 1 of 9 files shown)


alex-zaitsev:
## Final review — [#2046](https://github.com/Altinity/clickhouse-operator/pull/2046)

**Commits:** `93eaa5488` (shard safety) + `a1be5bb63` (operator IP restore). Branch matches remote.

### Verdict: approve

Addresses #1704 without dropping the operator-IP reconcile needed for `clickhouse-operator` user networks.

### Behavior (correct)

| Case | Outcome |
|------|---------|
| Generation same + healthy + IP same | Skip |
| Generation same + healthy + IP changed | Continue (networks refresh) |
| Unhealthy hosts | Continue (recovery) |
| Disruptive roll with no healthy peer | Defer host + `ErrCRUDAbort` |
| Unhealthy peer in shard | Recovery first; do not roll last healthy |

`prevCHOpIP` before `buildCR` is the right way to detect IP change.

### Residual nits (non-blocking)

1. Force-restart “skip” in `reconcileHostStatefulSet` only skips software restart; STS reconcile still runs. Shard-level deferral is the real gate.
2. Unused `down_start_time` in `test_010083`.
3. `hasUnhealthyHosts` does not short-circuit the host walk.

No further blockers from this pass.

Issue #2066, opened 2026-08-18
Reconcile of large CHIs is much slower since 0.27: ActionPlan is re-rendered on every status update


erikasb:
### What happened
 
After upgrading from 0.25.3 to 0.27.1, adding shards to a large (or even medium-sized) CHI became a lot slower (the larger the CHI, the more pronounced the difference).
Here are some comparisons from our test environment: one CHI with two clusters, scaling from 20 to 39 shards, 2 replicas each (80 -> 156 hosts total).
 
- 0.25.3: ~27 min
- 0.27.1: ~1h41m, operator CPU pinned at its limit for the whole run, and ~3000 "got error, will retry" status update conflict logs per run (vs ~250 on 0.25.3)
 
### Cause
After looking into, we identified the ActionPlan stringification to be the culprit.
Since the action plan was added to the status storage ConfigMap (0.27.x), buildStatusResourceData calls Status.ActionPlan.String() on every status update attempt. Each call re-renders the whole diff: it iterates the diff maps (random order, so the same plan produces different bytes every time) and Dump()s every diffed object — the full value is rendered and then truncated to 256 chars. For a scale-up the added items are entire shards, so every render dumps every added shard.
 
Status updates run several times per host, plus once per conflict retry, so that's thousands of renders per reconcile. In 0.25.x the plan was rendered exactly once per reconcile (LogActionPlan) and wasn't part of status writes at all.
 
It also feeds back on itself: the render makes each status update slower, which widens the window between reading the CR and writing the status, which causes more resourceVersion conflicts, which cause 1s-sleep retries, and every retry renders the plan again.
 
### Fix
 
Rendering the plan once at construction (see [PR](https://github.com/Altinity/clickhouse-operator/pull/2067)) restores 0.25.x behavior while keeping the storage feature. With that one change on top of 0.28.0, the same test scale-up takes 18–21 min, conflict retries are back to ~350, and operator CPU is back to 0.25.3 levels.

sunsingerus:
This PR may significantly help: https://github.com/Altinity/clickhouse-operator/pull/2067
Released just now, check please

PR #2067, opened 2026-08-18
Action plan render once

erikasb:
Fixes #[2066](https://github.com/Altinity/clickhouse-operator/issues/2066).

`ActionPlan.String()` is called on every status update attempt while building the status storage ConfigMap data. Each call re-renders the diff from scratch: reflection-dumps every diffed object (the full value is rendered, then truncated to 256 chars) and iterates the diff maps in random order, so the same plan produces different bytes on every call. During a scale-up of a large CHI this runs thousands of times per reconcile and dominates operator CPU (details and measurements in the issue).

This PR renders the plan once, eagerly in `MakeActionPlan`, and makes `String()` return that snapshot. The plan is immutable after construction, and every reconcile already renders it at least once for logging, so this adds no new work — it just stops repeating it. Before the plan was moved into the status path, it was rendered once per reconcile (`LogActionPlan`); this restores that frequency while keeping the storage feature.

Behavior notes:
- `String()` now returns the same bytes on every call. The stored `status-actionPlan` text becomes a construction-time snapshot. It is write-only (never parsed back), so nothing consumes the difference.
- Zero-value or JSON-unmarshaled plans (only unexported fields, so they unmarshal empty) render as `""`, same as before.

Measured on a 156-host CHI, scaling 2 clusters 20 → 39 shards each: 0.28.0 + this change completes in 18–21 min where 0.27.1 took ~1h41m with operator CPU pinned; status update conflict retries drop ~10x.


atsarevskiy:
Found this running your patch on top of 0.27.3: `render()` isn't safe on every
construction path, so moving the render into `MakeActionPlan` turns a latent nil read
into one that fires.

`specDiffReverse` is only assigned in the `(old != nil) && (new != nil)` branch, so the
nil-old, nil-new and both-nil paths leave it nil, and `render()` reads
`len(ap.specDiffReverse.Modified)` without a nil check. While `String()` was lazy that
read only happened where a caller actually wanted the string. Rendering at construction
time runs it on every path, so `MakeActionPlan(nil, cr)` panics.

Easiest way to see it is with #2070 also applied: its
`TestMarkReconcileStartReturnsStatusUpdateError` builds a plan with
`api.MakeActionPlan(nil, cr)`, and with this PR in the tree that test panics instead of
failing its assertion. Neither PR does it on its own, which is presumably why CI is
green here.

One line:

```go
if ap.specDiffReverse != nil && len(ap.specDiffReverse.Modified) > 0 {
```

Happy to push that as a commit if it's easier.

PR #2075, opened 2026-08-28
fix: use yaml 1.2 boolean format

marbon87:
Yaml 1.2 does not support yes/no anymore, use true/false instead

### DIFF
--- a/deploy/builder/templates-config/config.yaml
+++ b/deploy/builder/templates-config/config.yaml
@@ -489,9 +489,9 @@ reconcile:
       # replication delay a.k.a replication lag for the following replicas
       replicas:
         # All replicas (new and known earlier) are explicitly requested to wait for replication to catch-up
-        all: no
+        all: false
         # New replicas only are requested to wait for replication to catch-up
-        new: yes
+        new: true
         # Replication catch-up is considered to be completed as soon as replication delay
         # a.k.a replication lag - calculated as "MAX(absolute_delay) FROM system.replicas"
         # is within this specified delay (in seconds)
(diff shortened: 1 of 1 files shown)


sunsingerus:
Thanks for taking the time to send this, and sorry for the slow reply.

We looked into it and the premise doesn't hold for this codebase, so I'm going to close the PR — but the reasoning is worth recording, because it isn't obvious from the YAML alone.

**These settings are not plain booleans.** All seven are `*types.StringBool`, not `bool` — see `pkg/apis/clickhouse.altinity.com/v1/type_configuration_chop.go` (`ReconcileHostWaitReplicas.All`/`.New`, `ReconcileHostWaitProbes.Startup`/`.Readiness`, `ReconcileHostDropReplicas.OnDelete`/`.OnLostVolume`/`.Active`). `StringBool` is a string-backed type whose `IsTrue()`/`IsFalse()` accept `1/true/yes/on/enable/enabled` and `0/false/no/off/disable/disabled` (`pkg/apis/common/types/string_bool.go`). `yes`/`no` are supported by design, not by accident, and will keep working regardless of the YAML spec version.

**The config parser is YAML 1.1, not 1.2.** The operator config is loaded at `pkg/chop/config_manager.go:326` through `github.com/kubernetes-sigs/yaml`, which converts YAML to JSON via `gopkg.in/yaml.v2` — YAML 1.1 semantics, where `yes`/`no` already resolve to booleans.

We verified this end to end by feeding the shipped `config/config.yaml` through the operator's own loader, with and without this patch. The resulting values are identical:

```
                              current (yes/no)   this PR (true/false)
wait.replicas.all                  False               False
wait.replicas.new                  True                True
wait.probes.startup                False               False
wait.probes.readiness              True                True
drop.replicas.onDelete             True                True
drop.replicas.onLostVolume         True                True
drop.replicas.active               False               False
```

So there is no parse failure or misbehaviour to fix — the change is a no-op at runtime.

Two notes for anyone who finds this thread later:

- `deploy/builder/templates-config/config.yaml` is a **generator template**. Editing it alone is not enough: `dev/build_manifests.sh` and `dev/generate_helm_chart.sh` have to be re-run, or the template ends up disagreeing with nine committed artifacts while the shipped image keeps the previous values.
- There *is* a separate, genuine inconsistency in the Helm chart, where `values.schema.json` declares some of these as `type: boolean` while our own documentation examples use the quoted `"yes"` form. That is a Helm packaging matter and we will address it on its own.

Thanks again for the contribution.

